Issue #5, opened 2021-04-01
Consider ISO8601 for datetime which would be much easier to parse and validate across systems


jeffbaumes:
Basically the portal team would like to review with the metadata team the choice of this more difficult to parse datetime. If there are good reasons for this format we can keep it. If there are standard libraries for parsing these somewhat odd timestamps we should utilize them.

wdduncan:
I went through the schema looking for timestamp fields. In MIXS, I found the following fields took dates as values:
```
- fire
- last_clean
- date_last_rain
- iw_bt_date_well
- flooding
- extreme_event
- collection_date
```
Of these terms, only `collection_date` has data values associated with it, and these values are  already form  in `YYYY-MM-DD`.

As for the data that I am pulling in from GOLD, the only fields I found with datetime values were the `add_date` and `mod_date` fields in the `biosample` and `omics_processing` datasets. 

I also exact data for the `omics_processing.completion_date` field, but (currently) there is no data for this field.

wdduncan:
@dwinston I've made the changes to my code, but on the repo you moved the code that was in `metadata-translation/src/bin/lib`. I remember you mentioning something about this ... but where did you move it to?

Once we figure this out we can merge PR https://github.com/microbiomedata/nmdc-runtime/pull/6.

dwinston:
It is now in [metadata-translation/src/bin/lib](https://github.com/microbiomedata/nmdc-runtime/tree/2078951259f84a30784828663ddf2c82c00b02aa/metadata-translation/src/bin/lib) in the nmdc-runtime repo as well.

I had it there in my local repository, but it wasn't under version control because of a conflicting `lib/` entry for Python packaging in the `.gitignore`. I added `!metadata-translation/src/bin/lib` to the `.gitignore` to override this for our directory of interest.

You should be all set now. Let me know if anything else is blocking for you.

dwinston:
I merged in @wdduncan's PR #6. Will close this once the change is propagated for nmdc-server ingest, i.e. will co-close with #2.

wdduncan:
@ssarrafan this is pretty much finished, but waiting on @dwinston to close the ticket. So, I'll move this to the June sprint.

dwinston:
As the code for this is written, and what remains is for the metadata to be re-processed, I think this issue can be closed for now. It should be re-opened if any datetimes are discovered to be improperly encoded.

PR #6, opened 2021-05-21
Issue 5: convert dates to ISO-8601 format

wdduncan:
See #5


dwinston:
This still needs code added, right? Or is the PR intended to simply be an addition of example json files?

wdduncan:
@dwinston I've added the code to change the `add_date` and `mod_date` fields to iso-8601 format. I've also added the `nmdc_database.json.zip` that is the output of the `make run-etl` target. I wasn't sure if needed to re-run `make run-elt` to incorporate the new data or if you could slurp in the zip file.

PR #25, opened 2021-08-05
Issue 23: add functionality for hotfixes

wdduncan:
added simple file for working with dagster


dwinston:
Let's migrate the pipeline and solid code to `nmdc_runtime.pipelines.gold_translation` and `nmdc_runtime.solids.gold_translation`. You can migrate your `lib` modules to `nmdc_runtime.lib`.

dwinston:
The code migration from `metadata-translation` to `nmdc_runtime` modules should be sufficient to merge. After that, we'll want to:
(1) split logic into separate solids for emsl, gold, and jgi-other work.
(2) registering the necessary repo-resident data files (CSVs, etc.) as data repository service objects. This will enable us to get rid of the `metadata-translation` directory.

dwinston:
LGTM

Issue #28, opened 2021-08-25
Generate error reports for NMDC ETL staging


wdduncan:
Part of larger issue #23 

After data is loaded into the NMDC ETL staging database, we need to validate it against the JSON schema and return/out violations. These violations can then be addressed by the hotfixes.

cc @dwinston

ssarrafan:
@wdduncan is it ok to move this to the backlog?  or should it be moved to September's sprint?

wdduncan:
@ssarrafan   
Yes. Donny and I were working on this today, and are going to work on it some more tomorrow. We will finish this in the next sprint.

cmungall:
I think this is high priority

ssarrafan:
@dwinston and @wdduncan can this issue be closed?  I am assuming it's done but please confirm.

dwinston:
Yes, this can be closed.

Issue #23, opened 2021-07-28
Add hotfix functionality


wdduncan:
When the data needs to be changed, we need to implement a consistent method to make hotfixes to the data.

cc @cmungall @dehays @dwinston @wdduncan

wdduncan:
After meeting with @dwinston and @dehays, we decided on the attached image as a high level workflow for hotfixes.

![NMDC ETL Workflow Proposal](https://user-images.githubusercontent.com/3186638/128081634-f108b12f-0114-41d7-9b7d-c213de4143cb.png)

ssarrafan:
Added to August sprint and added Large label per Bill's update on the stand-up today.

ssarrafan:
@wdduncan since this is a large issue do you anticipate that it will still be done by the end of August?

wdduncan:
@ssarrafan I am working with @dwinston on this. I think we will be able to finish by the end of August.

ssarrafan:
Moving to Sept sprint per @dwinston

cmungall:
It looks like a lot of progress was made on the data model and serialization format for what we are calling the change sheets, in a google doc. Can than doc be linked here?

wdduncan:
Adding a link to Google doc spec of for the change sheet:  

https://docs.google.com/document/d/1gkz0Rp5vzX5KaOG749YIMmEpTM0ziVRvYQf4ZuYO3Ng/edit

(thanks for suggesting this @cmungall !)

wdduncan:
We need to determine how to add/update properties that are inlined and more than two levels deep. For example, the proposed way to relate a study to its team members:
```
study:
  ....
  slots:
  ...
     - has study role
 
study role:
  is_a: named thing
  aliases: ['credit table', 'associated researchers']
  ...
  slots:
    - applies to person
    - has contributor role
```
If we inline the study roles, the json would like this:
```json
{
  "id": "gold:Gs1234",
   "has study role": [
       {
           "applied to person": {"has raw value": "Jane Doe"},
           "has contributor role": "principal investigator"
        },
       {
           "applied to person": {"has raw value": "John Doe"},
           "has contributor role": "co-investigator"
       }
   ]
}
```
The problem is that it is unclear how to add/update a single study role. For example, suppose want to update Jane Doe's orcid:
|set|id|action|value|
|---|---|---|---|
|study|gold:Gs1234|has_study_role.applies_to_person.orcid| "..."|

[...]

PR #30, opened 2021-08-26
Issue 28: add schema validation to pipeline

wdduncan:
see issues #28  #23

### DIFF
--- a/nmdc_runtime/site/repository.py
+++ b/nmdc_runtime/site/repository.py
@@ -15,6 +15,7 @@ from starlette import status
 from toolz import merge
 
 from nmdc_runtime.api.core.util import dotted_path_for
+from nmdc_runtime.util import frozendict_recursive
 from nmdc_runtime.api.models.operation import ObjectPutMetadata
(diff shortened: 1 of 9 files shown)


dwinston (commented):
we're scheduled to meet soon, so we can clear things up then.

my general comment at this point is that I would like some statement of the goal, and some brief elaboration of strategy/tactics as appropriate, in the main conversation part of the PR, to guide my reading of the file diffs. This is because I'm not clear on what is being attempted here -- I guess our prior conversation has been booted from my working memory. sorry!

wdduncan:
@dwinston I added graph  for the emsl and gold validation.  
Since these graphs validate multiple data sets, I reverted using ops to get the collection names; e.g.:
```
@graph()
def gold():
    validate_mongo_collection(gold_study_set_collection_name())
    validate_mongo_collection(gold_omics_processing_set_collection_name())
    validate_mongo_collection(gold_biosample_set_collection_name())
```
Not sure how to pass multiple collection names using the config.  But we can revisit how to do it best.

PR #36, opened 2021-09-30
Issue 35: add SPRUCE data

wdduncan:
@dwinston @dehays  
I have made the necessary changes to include projects that don't have outputs. However, I am not confident in the counts.

In my gold dump, there are a total 889 projects. Of these:
- 168 have data objects as outputs.
- There are 123 projects in SPRUCE study

This leaves 598 projects in the exported data that do not have outputs and are not in the SPRUCE study. Do we still want to include those projects?

### DIFF
--- a/metadata-translation/src/bin/lib/nmdc_data_source.yaml
+++ b/metadata-translation/src/bin/lib/nmdc_data_source.yaml
@@ -151,7 +151,6 @@ classes:
       description: description
       part_of: [study_gold_id]
       has_input: [biosample_gold_id]
-      has_output: [output_file_ids]
     transforms:
       post:
         # note: attribute names must match names in nmdc.yaml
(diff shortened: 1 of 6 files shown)


dehays:
@wdduncan. I think what you are seeing are many sequencing projects (from the other 2 SFA studies and from the BioScales study) that we do not want to include because they have no NMDC analysis.  

We need a process by which projects are included when there is analysis.  As we have discussed, that falls on the boundary between automated inclusion of analysis and the logic for the ETL including projects and setting their has_output.

This week - I suggest constraining to only include the SPRUCE study projects.

wdduncan:
@dehays I can filter out those other projects (i.e., those not having an output and not being included in SPRUCE). The drawback (of course) is that makes the etl more ad hoc.

PR #64, opened 2021-11-01
DO NOT MERGE: test experiment

cmungall:
@wdduncan and I are attempting to replicate test environment locally, this is just an experiment to check what gh actions is doing


cmungall:
Confirmed: no gh actions

dwinston:
Correctly failed!

PR #69, opened 2021-11-03
Null change to test actions

cmungall:


dwinston:
@cmungall @elais the main branch is in a failing state wrt GitHub actions because the GitHub action yaml does not yet properly set up the environment for testing (e.g. env vars supplied via GitHub secrets, docker compose to start the needed services like mongo, etc.). I had only been running tests before prior to pushing to main. A bit more work is needed for this PR to check green.

dwinston:
@cmungall @elais tests are running properly now via github actions.

PR #74, opened 2021-11-11
issue-31: create delimited list for multiple values in change sheets

### LINKED ISSUE #31: create delimited list for multiple values in change sheets
Related to https://github.com/microbiomedata/nmdc-schema/issues/159 

When entering a list of values into an array in the change sheet, use a delimiter (e.g. `|`) insert multiple values.

cc @dehays

wdduncan:
Fixes #31 

Starting to work on this. Currently, I just updated with docstrings and typing.


dwinston (commented):
Looks fine so far. Thanks for adding the type hints and docstrings.

elais:
Is this ready to be merged?

wdduncan:
@elais Thanks for checking, but no, it is not ready. When it is ready, I will change the status from "Draft" to "Ready for review".

elais:
Sounds good. When you do could you make sure to rebase on top of the main branch and squash the commits? That way we can have only one log entry tied to this.

cmungall:
It is failing on this:

```python
tests/test_util.py:11: in <module>
    from nmdc_schema.nmdc_data import get_nmdc_jsonschema_dict
E   ModuleNotFoundError: No module named 'nmdc_schema.nmdc_data'
```

This is dependent on: https://github.com/microbiomedata/nmdc-schema/pull/211 which was merged 26 days ago

but nmdc-runtime is pinned to an earlier version:

https://github.com/microbiomedata/nmdc-runtime/runs/4527481257?check_suite_focus=true#step:6:1736

```
Collecting nmdc-schema==2021.11.15rc1
  Downloading nmdc_schema-2021.11.15rc1-py3-none-any.whl (135 kB)
```

so it's not surprising this PR fails

The latest nmdc_schema is 2021.11.22rc1

this code works with this latest version:

```python
>>> from nmdc_schema.nmdc_data import get_nmdc_jsonschema_dict
```

cmungall:
I see it's failing with

```
    def opt_dict_param(
        obj: Any,
        param_name: str,
        key_type: Type = None,
        value_type: Type = None,
        value_class: Type = None,
    ) -> Dict:
        """Ensures argument obj is either a dictionary or None; if the latter, instantiates an empty
        dictionary.
        """
        from dagster.utils import frozendict
    
        if obj is not None and not isinstance(obj, (frozendict, dict)):
>           raise _param_type_mismatch_exception(obj, (frozendict, dict), param_name)
E           dagster.check.ParameterCheckError: Param "config" is not one of ['dict', 'frozendict']. Got frozendict.frozendict({'host': frozendict.frozendict({'env': 'MONGO_HOST'}), 'username': frozendict.frozendict({'env': 'MONGO_USERNAME'}), 'password': frozendict.frozendict({'env': 'MONGO_PASSWORD'}), 'dbname': frozendict.frozendict({'env': 'MONGO_DBNAME'})}) which is type <class 'frozendict.frozendict'>.
```

this change seems unrelated to any changes you have made. As a sanity check I would just check that master/main actually succeeds. I might then investigate whether the cause is to do with bringing in a new library - make a PR that only incorporates the library/dependency changes and see if that recapitulates the problem.

If you look at the code you can see what the problem is: dagster is expecting a `dagset.utils.frozendict` object, whereas util.py is using the standard pypi https://pypi.org/project/frozendict/ library. Despite the fact the name is the same, **these are different classes as far as python is concerned**.

What is not clear to me is how this even worked in the first place! I really don't know much about dagster and why we're feeding it frozendicts, we may need Donny or a dagster expert to shine some light on what util.py does.

My best guess is that something changed in the dagster core here, and previous PRs worked against the old version, and (unrelated to your changes), it no longer works, but that is a guess

PR #92, opened 2022-01-26
feat: update-schema make target

dwinston:
Can now `make update-schema` to only `--upgrade-package nmdc-schema` (rather than a broader `--upgrade`) when `pip-compile`ing deps to `requirements/main.txt`.

\cc @wdduncan @scanon @dehays

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,10 @@ update-deps:
 
 update: update-deps init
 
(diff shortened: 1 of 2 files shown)


dwinston:
manually ran tests. but also changed gh action to trigger py tests on deps (requirements/main.txt) updates: 799a644e3edc65e44c520a8bbdb5606b78cb5005

PR #106, opened 2022-03-09
feat(endpoints): add example task endpoint

elais:
The example job endpoint is for example only, demonstrating a workflow
that actively runs jobs given a user's call to a single endpoint (rather
than multiple calls to multiple endpoints) using celery and
mongodb. Effort was taken to minimize the impact on existing code though
integration in the future would be straightforward.

A few key points:

- The API server handles both supplying jobs to the celery worker and
storing the results in the database

- This is meant to be run on each call as opposed to passively as is the
case with Dagster.

- Shouldn't be seen as a replacement for existing (and well documented)
infrastructure, but an alternative to exist alongside it.

- Future work would involve further decoupling of models from the api
endpoints and maybe deeper thought into what endpoint to put future
services like this under (like, `/services` for instance).


dwinston:
@elais have a look at <https://docs.dagster.io/concepts/dagit/graphql#launch-a-run>. I think we may be able to POST to the dagster graphql API in the body of a FastAPI route handler, get back a run id, and set up another FastAPI route to get status.

elais:
How does nmdc-runtime access remote file systems, and what needs to be done to make sure it queries the correct file system when looking to upload files? Right now the code in the pull request expects the file to be in a local filesystem. I see some code to leverage s3, should we expect remote files to come from a bucket or try to parse inputs to see which backend to use?

dwinston:
@elais the current mechanism for remote file access is via the DRS API (`/objects` route family).

For example, the [`get_json_in`](https://github.com/microbiomedata/nmdc-runtime/blob/358ff4383ef34a10fb22b448b59f372187c46c66/nmdc_runtime/site/ops.py#L413) op uses a DRS object_id and calls [`get_object_bytes`](https://github.com/microbiomedata/nmdc-runtime/blob/49b03d05e49c4c6e2470696d405df5c5e44bc376/nmdc_runtime/site/resources.py#L120) to resolve a URL (even for S3, a presigned URL is generated on the fly on request), but the DRS API can encode access mechanisms other than https. This is not implemented, though.

dwinston:
The current NMDC DataObject schema requires a "url" field, so supporting remote filesystems without a fronting http proxy has not been implemented in production.

elais:
> The current NMDC DataObject schema requires a "url" field, so supporting remote filesystems without a fronting http proxy has not been implemented in production.

That's good. I talked with the LANL team about using a reverse proxy to make our files available (would be great for testing).

dwinston:
@elais this has been superseded by related work and can be closed, right?

elais:
Right, this pull request can be closed.

PR #107, opened 2022-03-18
feat: /runs endpoints for interactive job-run requests/monitoring

dwinston:
Using `DagsterGraphQLClient` under the hood. Simple example job given.


nmdc_runtime/api/endpoints/runs.py
```diff
@@ -0,0 +1,51 @@
...
+from fastapi import APIRouter
+from starlette import status
+
+router = APIRouter()
+
+DAGIT_HOSTNAME, DAGIT_PORT = os.getenv("DAGIT_HOST").split("://", 1)[-1].split(":", 1)
+DAGIT_PORT = int(DAGIT_PORT)
+
+
+@router.post("/runs:request")
```
elais, line 15: What if this were structured as a GET to `/runs/<job-name>` with the config data structured as a query like `q='{"ops":{"hello":"beef}}`? That way we can avoid POST calls that semantically should map to creating objects and use GET calls that semantically should map to read calls. We could also build the query in a way that will return a collection, leaving it flexible for when we want to pull multiple runs (and saving some code in the process)

dwinston: I understand and agree with your reasoning but not your conclusions. I agree that POST calls generally map to creating things and GET calls map to no-side-effect reads. In this case, the API user is requesting a new job run to be launched, and gets in response the ID of the new run. So...POST seems appropriate vs GET, right?

nmdc_runtime/api/endpoints/runs.py
```diff
@@ -0,0 +1,51 @@
...
+            run_config=run_config_data,
+        )
+        return {"new_run_id": new_run_id}
+    except DagsterGraphQLClientError as exc:
+        raise HTTPException(
+            status_code=status.HTTP_412_PRECONDITION_FAILED, detail=str(exc)
+        )
+
+
+@router.get("/runs/{run_id}/status")
```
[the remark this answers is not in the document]
dwinston: Sure, that's fine.

nmdc_runtime/api/main.py
```diff
@@ -21,6 +21,7 @@
     queries,
     metadata,
     nmdcschema,
+    runs,
 )
 from nmdc_runtime.api.models.site import SiteInDB, SiteClientInDB
```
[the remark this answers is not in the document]
dwinston: that's fair. we can do that.

nmdc_runtime/api/endpoints/runs.py
elais: I think I misread this. You're right post is appropriate. Would we want to use an array for the payload so users can start one or more jobs?

PR #156, opened 2022-09-29
migration of ID minting to include typecodes

dwinston:


dwinston:
note: this is NOT ready to merge. just sharing progress.

turbomam:
> I had a preliminary discussion of the below diagram with @turbomam . Hoping to have a semi-populated model (i.e. with owl:sameAs instance IDs currently in the metadata store in various mongo collections).

I like this diagram, but I don't believe it's normative to the nmdc-schema. Are you suggesting that we add more classes?

dwinston:
> I like this diagram, but I don't believe it's normative to the nmdc-schema. Are you suggesting that we add more classes?
I guess I might be, but that is out of scope for current squad work. I don't plan to pursue it.

dwinston:
superceded by #186

PR #166, opened 2022-10-27
feat(outputs): add client site authentication

elais:
feat(api/jobs): return workflow

Instead of job specific data return the workflow record when a claim is made. Compute automation needs the workflow object returned to continue their happy path.

feat(releases): bump release


Michal-Babins (approved):
Changes to return job claim to kick off next job.

PR #174, opened 2022-12-22
Add ingest pipeline with GOLD API source

pkalita-lbl:
The goal of these changes is to establish patterns for brining data from external sources via the Dagster infrastructure. To this point a lot of that work has been done in a [separate repo](https://github.com/microbiomedata/sample-annotator), and a lot of it has been focused on accessing data from GOLD. These changes add classes for accessing data from the GOLD API (`GoldApiClient`) and transforming it into NMDC schema-compliant objects (`GoldTranslator`). It also wraps calls to those classes in Dagster `@op`s and `@graph`s.

PR #176, opened 2023-01-09
add `@op` method to compute study_set slot

sujaypatil96:
This PR seeks to build on top of PR #174 

The practice task that I'm trying to accomplish as part of this PR is the addition of a method to compute the study_set slot for an NMDC Database object.


cmungall:
I'm a little unclear on the division of responsibility of this repo and sample-annotator when it comes to GOLD ingests

pkalita-lbl:
My perspective is that this work will supersede sample-annotator (or at least part of it, as sample-annotator does a few different things). I think of it like sample-annotator is the prototype version and this will be the production version. We'll lift over whatever pieces of sample-annotator are useful, modifying as needed, because an important distinction is that sample-annotator attempts to produce data instances that adhere to [nmdc-schema](https://github.com/microbiomedata/nmdc-schema) whereas the thought with the Biosample Ingest Squad, in general, is to produce data formatted for the submission portal (which is only partially covered by a schema, but that's a separate issue). 

What might be causing a little confusion is that these changes _are_ producing nmdc-schema instances. I think this was just Sujay getting setup with this project and wasn't intended to be final code.

If any of that sounds concerning to you I'd be happy to chat more!

PR #186, opened 2023-01-19
/pids/* to mint and manage draft IDs

dwinston:
This is not feature-complete, but good to demo and get UX feedback. It mints draft IDs (persisted but not yet resolvable).


turbomam:
Great! Where are you getting the typecodes from? I'm adding them from [20230112 nmdc typecodes ](https://docs.google.com/spreadsheets/d/17m7rtf66ffMN2VwIfO3nZyJ0f-4BgXngMwI43XWsHb4/edit#gid=0) into `slot_usage` statements now.

Relevant PR:
- https://github.com/microbiomedata/nmdc-schema/pull/616/files

Depending on which of my posts or code samples you've read before, you will probably see some new implementation. We will directly assigning a typecode literal, rather than using the symbolic `settings` like `{id_typecode_activity}`

I am also adding literal `pattern`s in addition to the `structured_pattern`s as I don't believe the `structured_pattern` expander is working yet. As soon as it does, I will remove the `pattern`s.

dwinston:
tests pass locally; not sure why CI breaks on `len(rv) == 1 and rv[0].startswith("nmdc:")` with a `KeyError` on `0`.

elais:
We should really look into this. I get sometimey test failures that I believe are happening do some async call.

PR #205, opened 2023-02-01
Makefile rule for getting submission schema and its artifacts

### LINKED ISSUE #190: Generate Python classes from submission schema
For biosample metadata ingest pipelines we want to produce data artifacts that can be picked up by the submission portal. To make the pipeline code less error-prone it would be useful to have Python classes generated from the submission schema. Currently this schema is generated as an artifact of https://github.com/microbiomedata/sheets_and_friends. The simplest solution might be to add a Makefile target that fetches the schema submission YAML-file from a specific revision of sheets_and_friends and use the LinkML CLI to generate Python classes from it. The other option is to have sheets_and_friends do the generation and publish the result to PyPI. That might be better in the long run, but also might take a little longer to get set up.

sujaypatil96:
TODO:
- [x] Get raw YAML file from tag and not the `main` branch specifically

Fixes #190 

CC: @pkalita-lbl

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,5 +94,14 @@ mongorestore-nmdcdb-lite-archive:
 quick-blade:
 	python -c "from nmdc_runtime.api.core.idgen import generate_id; print(f'nmdc:nt-11-{generate_id(length=8, split_every=0)}')"
 
+# NMDC Submission Portal schema Makefile rules
+SCHEMA_DIR = schema
+SCHEMA_NAME = nmdc_submission_schema
+SCHEMA_SRC = $(SCHEMA_DIR)/$(SCHEMA_NAME).yaml
+submission-schema-artifacts:
+	mkdir -p $(SCHEMA_DIR)
+	curl -L https://raw.githubusercontent.com/microbiomedata/sheets_and_friends/main/artifacts/nmdc_submission_schema.yaml > $(SCHEMA_SRC)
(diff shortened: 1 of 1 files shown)


sujaypatil96:
Can be used successfully on resolution of https://github.com/microbiomedata/sheets_and_friends/issues/165

dwinston (commented):
Seems like the whole URL to the yaml can be set as an env var. I don't see the benefit of breaking up into separate variables at this stage with only the one Make target.

Also, it seems like this could be integrated with the nmdc-schema build and thus distributed with the PyPI package?

pkalita-lbl:
I'll justify this because I suggested doing it this way in the first place 😁 Admittedly it's a little wonky, but it's precisely because (as you rightly pointed out) the submission schema is decoupled from `nmdc-schema` and also doesn't have it's own standard build and publish process. Ideally either the submission schema would come with the `nmdc-schema` dependency or a new dependency. But that's a considerably bigger task and I didn't want to block other work on that.

Since this is intended to make the submission schema an ersatz dependency, I would argue against having the URL to the YAML in an environment variable. If it were in an environment variable, the version of the submission schema dependency would be decoupled from version of the code using it (kind of like a `requirements.txt` file with no version specifications 😁)

I had envisioned that the artifact produced by the Makefile target would be checked in though. That way developers don't have to remember to run the target unless they're specifically upgrading the version.

PR #210, opened 2023-02-03
Add taxonomy to read based analysis set name

elais:


components/nmdc_runtime/workflow/spec.py
```diff
@@ -151,8 +151,8 @@ class ReadBasedAnalysis(Workflow):
     version: str = "1.0.2"
```
turbomam, line 151: If you change a collection name in MongoDB, a corresponding change will be required in the schema. Could you please open an issue for that in nmdc-schema?

elais: Hi Mark, 

I mentioned this on slack but for posterity's sake I just want to assure you the original name `read_based_analysis_activity_set` was _not_ schema compliant and this update is a bug fix to make it so by renaming the collection `read_based_taxonomy_analysis_activity_set`. There aren't any changes needed in the schema for this.

Issue #240, opened 2023-03-24
API validation to make sure nmdc identifiers are not duplicated


aclum:
the value for 'id' in mongo should be unique. The APi should validate that submissions don't contain an identifier that has already been used. Example: 
A submission was allowed with duplicate values for 'id' for field_research_site_set
{
    "_id" : ObjectId("641dde3711db7ba4959f2d2b"),
    "id" : "nmdc:frsite-11-s2dqk408",
    "name" : "BESC-470-CL2_38_23"
}
{
    "_id" : ObjectId("641dde3711db7ba4959f2d88"),
    "id" : "nmdc:frsite-11-s2dqk408",
    "name" : "BESC-470-CL2_38_23"
}
{
    "_id" : ObjectId("641dde3711db7ba4959f2e13"),
    "id" : "nmdc:frsite-11-s2dqk408",
    "name" : "BESC-470-CL2_38_23"
}

dwinston:
@aclum how can this situation be reproduced? I see that there are no duplicates now. Was this situation created via direct mongo commands? Or via an API endpoint?

dwinston:
Okay, I see what happened. The logic for ensuring unique indexes depends on detection of a document with an "id" field being present in a non-`system.` collection: https://github.com/microbiomedata/nmdc-runtime/blob/0a79872e1f5e562be7d25be726d0aea224a46389/nmdc_runtime/api/main.py#L325

However, this means that a new (empty) collection added to the schema, e.g. `field_research_site_set` is a recent addition, would not have a unique index set for it until after some documents are added. So, one can add a set of documents with shared IDs and it not be rejected by mongodb as an index integrity violation.

To guard against this, I will

- (a) ensure full simulation of database action during json validation (this is done for changesheets:validate, i.e. a temporary mongodb database is created for each validation request, but not currently for json:validate)
- (b) change the ensure-unique-indexes logic to be sourced from the schema (either directly from the json schema per https://github.com/microbiomedata/nmdc-schema/issues/852 or otherwise pre-computed).

dwinston:
#245 addresses part (b) above.

PR #245, opened 2023-04-17
fix: ensuring indexes on 'id'

dwinston:
source from nmdc schema and compatible non-system collections

part of #240

### DIFF
--- a/nmdc_runtime/api/core/metadata.py
+++ b/nmdc_runtime/api/core/metadata.py
@@ -37,11 +37,14 @@ collection_name_to_class_name = {
     if "items" in db_prop_spec and "$ref" in db_prop_spec["items"]
 }
 
-collection_names_with_unique_id = {
-    coll_name
-    for coll_name, class_name in collection_name_to_class_name.items()
-    if "id" in get_nmdc_jsonschema_dict()["$defs"][class_name].get("required", [])
-}
+
+@lru_cache
+def schema_collection_names_with_id_field():
+    return {
+        coll_name
+        for coll_name, class_name in collection_name_to_class_name.items()
+        if "id" in get_nmdc_jsonschema_dict()["$defs"][class_name].get("properties", {})
+    }
 
 
 def load_changesheet(
--- a/nmdc_runtime/api/db/mongo.py
+++ b/nmdc_runtime/api/db/mongo.py
@@ -1,18 +1,23 @@
 import os
+from contextlib import AbstractContextManager
 from functools import lru_cache
-from typing import Set
+from typing import Set, Dict, Any, Iterable
+from uuid import uuid4
 
-import pymongo.errors
+from pymongo.errors import OperationFailure, AutoReconnect
 from motor.motor_asyncio import AsyncIOMotorClient, AsyncIOMotorDatabase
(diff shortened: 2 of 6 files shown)


nmdc_runtime/api/core/metadata.py
```diff
@@ -37,11 +37,14 @@
     if "items" in db_prop_spec and "$ref" in db_prop_spec["items"]
 }
 
-collection_names_with_unique_id = {
-    coll_name
-    for coll_name, class_name in collection_name_to_class_name.items()
-    if "id" in get_nmdc_jsonschema_dict()["$defs"][class_name].get("required", [])
-}
+
+@lru_cache
```
shreddd, line 41: Do we need to worry about cache invalidation? Or would  that only happen if a new version is deployed.

nmdc_runtime/api/db/mongo.py
```diff
@@ -73,3 +78,107 @@ def activity_collection_names(mdb: MongoDatabase) -> Set[str]:
...
+    return len(coll.distinct("id")) == total_count
+
+
+def ensure_unique_id_indexes(mdb: MongoDatabase):
+    """Ensure that any collections with an "id" field have an index on "id"."""
+    candidate_names = (
+        set(mdb.list_collection_names()) | schema_collection_names_with_id_field()
+    )
+    for collection_name in candidate_names:
+        if collection_name.startswith("system."):  # reserved by mongodb
```
shreddd, line 99: Should we also ignore "internal" NMDC collections beginning with _

nmdc_runtime/api/db/mongo.py
```diff
@@ -73,3 +78,107 @@ def activity_collection_names(mdb: MongoDatabase) -> Set[str]:
...
+class DeleteStatement(BaseModel):
+    q: dict
+    limit: conint(ge=0, le=1) = 1
+
+
+class OverlayDBError(Exception):
+    pass
+
+
+class OverlayDB(AbstractContextManager):
```
shreddd, line 125: Can you add a couple of comments explaining what the OverlayDB class is for

nmdc_runtime/api/core/metadata.py
dwinston: right, we explicitly update the `nmdc-schema` PyPI package dependency, and that necessitates a fresh deployment, so `get_nmdc_jsonschema_dict()` can be cached on init for a deployment.

nmdc_runtime/api/db/mongo.py
dwinston: Interesting. I'd like to not ignore them, and lean on the semantic that a collection with at least one document, an "id" field in every document, and all "id" values different, i.e. what is checked by `all_docs_have_unique_id`, should ensure a unique index on "id".

I can be swayed to also ignore "_"-beginning collection names if you want that to approve this PR. The current behavior is inessential.

nmdc_runtime/api/db/mongo.py
dwinston: Done.

shreddd (approved):
OK - this makes sense. Ready to merge.

PR #255, opened 2023-06-13
Use samp_name slot from submission portal to also populate name of Biosample

pkalita-lbl:
Just a small fix for something that was initially missed when working on https://github.com/microbiomedata/issues/issues/224

### DIFF
--- a/nmdc_runtime/site/translation/submission_portal_translator.py
+++ b/nmdc_runtime/site/translation/submission_portal_translator.py
@@ -382,7 +382,11 @@ class SubmissionPortalTranslator(Translator):
         :param nmdc_study_id: Minted nmdc:Study identifier for the related Study
         :return: nmdc:Biosample
         """
-        slots = {"id": nmdc_biosample_id, "part_of": nmdc_study_id}
+        slots = {
+            "id": nmdc_biosample_id,
+            "part_of": nmdc_study_id,
+            "name": sample_data[0].get("samp_name"),
+        }
         biosample_slot_names = self.schema_view.class_slots("Biosample")
         for tab in sample_data:
             for column, value in tab.items():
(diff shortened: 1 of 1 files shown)


pkalita-lbl:
Paging @dwinston

nmdc_runtime/site/translation/submission_portal_translator.py
```diff
@@ -382,7 +382,11 @@ def _translate_biosample(
         :param nmdc_study_id: Minted nmdc:Study identifier for the related Study
         :return: nmdc:Biosample
         """
-        slots = {"id": nmdc_biosample_id, "part_of": nmdc_study_id}
+        slots = {
+            "id": nmdc_biosample_id,
+            "part_of": nmdc_study_id,
+            "name": sample_data[0].get("samp_name"),
```
dwinston, line 388: is `sample_data` guaranteed to be non-empty? and is the first element of `sample_data` guaranteed to have a "samp_name" key? If no to the latter, is `None` acceptable to assign to "name"?

pkalita-lbl: > is sample_data guaranteed to be non-empty?

[Yes](https://github.com/microbiomedata/nmdc-runtime/blob/c3c2af2039c32a3b1d63cb04c838ab80e2c81cb2/nmdc_runtime/site/translation/submission_portal_translator.py#L444)

> is the first element of sample_data guaranteed to have a "samp_name" key? 

[Yes](https://github.com/microbiomedata/submission-schema/blob/main/src/nmdc_submission_schema/schema/nmdc_submission_schema.yaml#L46282)

> is None acceptable to assign to "name"

[Yes](https://github.com/microbiomedata/nmdc-schema/blob/ebfd7dd1cf25f15f5d95f96ac3ff9a113f5d81c9/src/schema/basic_slots.yaml#L37-L40)

dwinston (approved):

PR #275, opened 2023-07-14
Update main.in

shreddd:


dwinston:
@turbomam is any data migration required? I read https://github.com/microbiomedata/nmdc-schema/releases/tag/v7.7.2 and it's not apparent. By semvar convention, this is a patch release, but I'm not sure if this corresponds to "no db migration needed" in our case.

dwinston:
per conversation with @shreddd , this schema update shouldn’t require data migration.

PR #289, opened 2023-08-21
Implement projection for the API 

### LINKED ISSUE #280: Implement projection for the API
Per the [conversation on Slack](https://nmdc-group.slack.com/archives/C02KM21QJ4E/p1690746669441729?thread_ts=1689697607.364339&cid=C02KM21QJ4E) with @dwinston . It would be helpful to make requests like, get all the data from a particular study. Along the lines:

`get_json("/nmdcschema/study_set/nmdc:sty-11-5tgfr349").get("ess_dive_datasets")`

dwinston:
feat: mongodb projection for list_resources/ListRequest

closes #280

### DIFF
--- a/nmdc_runtime/api/endpoints/util.py
+++ b/nmdc_runtime/api/endpoints/util.py
@@ -60,6 +60,7 @@ HOSTNAME_EXTERNAL = BASE_URL_EXTERNAL.split("://", 1)[-1]
 def list_resources(req: ListRequest, mdb: MongoDatabase, collection_name: str):
     limit = req.max_page_size
     filter_ = json_util.loads(req.filter) if req.filter else {}
+    projection = req.projection.split(",") if req.projection else None
     if req.page_token:
         doc = mdb.page_tokens.find_one({"_id": req.page_token, "ns": collection_name})
         if doc is None:
@@ -77,7 +78,11 @@ def list_resources(req: ListRequest, mdb: MongoDatabase, collection_name: str):
             filter_ = merge(filter_, {"id": {"$gt": last_id}})
 
     if mdb[collection_name].count_documents(filter=filter_) <= limit:
-        rv = {"resources": list(mdb[collection_name].find(filter=filter_))}
+        rv = {
+            "resources": list(
+                mdb[collection_name].find(filter=filter_, projection=projection)
+            )
+        }
         return rv
     else:
         if "id_1" not in mdb[collection_name].index_information():
@@ -86,7 +91,11 @@ def list_resources(req: ListRequest, mdb: MongoDatabase, collection_name: str):
             )
         resources = list(
             mdb[collection_name].find(
-                filter=filter_, limit=limit, sort=[("id", 1)], allow_disk_use=True
+                filter=filter_,
+                projection=projection,
+                limit=limit,
+                sort=[("id", 1)],
+                allow_disk_use=True,
             )
         )
         last_id = resources[-1]["id"]
--- a/nmdc_runtime/api/models/util.py
+++ b/nmdc_runtime/api/models/util.py
@@ -24,7 +24,9 @@ class ListRequest(BaseModel):
     page_token: Optional[str]
     projection: Annotated[
         Optional[str],
-        Query(description="comma-separated field names. Example: `id,doi`"),
+        Query(
+            description="for MongoDB-like [projection](https://www.mongodb.com/docs/manual/tutorial/project-fields-from-query-results/): comma-separated list of fields you want the objects in the response to include. Example: `id,doi`"
+        ),
     ]
 
 
--- a/nmdc_runtime/api/models/util.py
+++ b/nmdc_runtime/api/models/util.py
@@ -1,4 +1,6 @@
-from typing import TypeVar, List, Optional, Generic
+from typing import TypeVar, List, Optional, Generic, Annotated
+
+from fastapi import Query
 
 from pydantic import BaseModel, root_validator, conint
 from pydantic.generics import GenericModel
@@ -12,9 +14,18 @@ class ListResponse(GenericModel, Generic[ResultT]):
 
 
 class ListRequest(BaseModel):
-    filter: Optional[str]
+    filter: Annotated[
+        Optional[str],
+        Query(
+            description='MongoDB-style JSON filter document. Example: `{"ecosystem_type": "Freshwater"}`'
+        ),
+    ]
     max_page_size: Optional[int] = 20
     page_token: Optional[str]
+    projection: Annotated[
+        Optional[str],
+        Query(description="comma-separated field names. Example: `id,doi`"),
+    ]
 
 
 PerPageRange = conint(gt=0, le=200)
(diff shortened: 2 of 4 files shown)


nmdc_runtime/api/endpoints/util.py
```diff
@@ -60,6 +60,7 @@
 def list_resources(req: ListRequest, mdb: MongoDatabase, collection_name: str):
     limit = req.max_page_size
     filter_ = json_util.loads(req.filter) if req.filter else {}
+    projection = req.projection.split(",") if req.projection else None
```
eecavanna, line 63: Note to self (and other reviewers):
- According to [the `find` docs](https://pymongo.readthedocs.io/en/stable/api/pymongo/collection.html#pymongo.collection.Collection.find), `None` is the default value of the `projection` parameter; so, when the client (i.e. request) doesn't specify a projection list, the behavior of `find` will be like it was before this change to the code.

nmdc_runtime/api/models/util.py
```diff
@@ -12,9 +14,18 @@ class ListResponse(GenericModel, Generic[ResultT]):
...
+        Optional[str],
+        Query(
+            description='MongoDB-style JSON filter document. Example: `{"ecosystem_type": "Freshwater"}`'
+        ),
+    ]
     max_page_size: Optional[int] = 20
     page_token: Optional[str]
+    projection: Annotated[
+        Optional[str],
+        Query(description="comma-separated field names. Example: `id,doi`"),
```
eecavanna, line 27: I didn't know what "projection" meant in this context, but, once I read about it in the pymongo and MongoDB docs, I understood what it meant. Based on that experience, I have a suggestion for this description:
```diff
- comma-separated field names.
+ comma-separated list of field names you want the objects in the response to include.
```

eecavanna (commented):
LGTM - acting upon the suggestion about the annotation/parameter description is optional from my perspective.

brynnz22:
Thanks so much for implementing this!

brynnz22:
Sorry, I am late reviewing this. What are the restrictions for the projection parameter? I was testing it using the GUI with  `study_set` , which seems to work great for all the various study fields, but when I test it with `biosample_set`  it only seems to work for the `id` field, but not others like `ecosystem_type`. It returns an Internal Server Error.

eecavanna:
I am seeing what @brynnz22 reported [here](https://github.com/microbiomedata/nmdc-runtime/pull/289#issuecomment-1687106658).

Case 1: `collection_name` is `study_set` and `projection` is `ecosystem_type` -> HTTP 200

image

Case 2: `collection_name` is `biosample_set ` and `projection` is `ecosystem_type` -> HTTP 500

image

---

I also tried with a field name that doesn't exist in either collection (i.e. the field name, `foo`):

Case 3: `collection_name` is `study_set` and `projection` is `foo` -> HTTP 200

image

Case 4: `collection_name` is `biosample_set ` and `projection` is `foo` -> HTTP 500

image

eecavanna:
Based on what I see in the [server logs](https://github.com/microbiomedata/nmdc-runtime/pull/289#issuecomment-1687129248), I think the `list_resources` function defined in `nmdc_runtime/api/endpoints/util.py` is trying to access the `id` key of a dictionary that doesn't have an `id` key (maybe because the projection omits it). I wonder if the projection logic could be updated to always include the `id` key (if an projection value is defined in the request).

CC: @dwinston

eecavanna:
I just noticed something about Cases 2 and 4, which are defined [in this earlier comment](https://github.com/microbiomedata/nmdc-runtime/pull/289#issuecomment-1687129248) and, as-is, result in HTTP 500 responses.

If I add `id` to the projections list, the response is an HTTP 200 response.

Case 2B:

image

Case 4B:

image

CC: @dwinston , @brynnz22

eecavanna:
What I don't understand is why explicitly including `id` seems to be necessary for the `biosample_set` collection, but not for the `study_set` collection.

eecavanna:
> if there are enough documents in that collection that the else block gets run, the system hits the issue

I think that's it.

**Case 5A:**
- Page size = 17
- Number of documents in collection = 18
- `id` is not in the projections list

Response is HTTP 500 🔴 

image

**Case 5B:**
- Page size = 18
- Number of documents in collection = 18 (same as Case 5A)
- `id` is not in the projections list (same as Case 5A)

Response is HTTP 200 🟢 

image

brynnz22:
@eecavanna thanks for looking into it! So does the user need to make the max_page_size bigger than the number of documents?

Something I also noticed, which is very minor, is that using the GUI, the projections are comma-separated with no space. If I put a space between the projections, it will only return the first field specified. For example for `study_set` with a projection of `name,ecosystem_type`, it will only include the `name` in the results if the projection is typed as `name, ecosystem_type`. Can it be adjusted to account for white spaces? That would be consistent with how some of the other endpoints are structured. It looks like calling the [get_pairs](https://github.com/microbiomedata/nmdc-runtime/blob/0363d81d2a355c2186df6efff6ee1742eaba3772/nmdc_runtime/api/endpoints/util.py#L116) function can be used to do this?

eecavanna:
Hi @brynnz22, thanks for reporting the original issue!

>So does the user need to make the `max_page_size` bigger than the number of documents?

That is one workaround. The other is to—if specifying any projections—include `id` in the list.

I think the behavior you reported is a bug (i.e. not by design) and will eventually be fixed; but, in the meantime, doing either of those two things will work around the issue.

I'll respond to the other paragraph in your comment separately.

eecavanna:
> If I put a space between the projections, it will only return the first field specified.

The [code that consumes the projections parameter value](https://github.com/microbiomedata/nmdc-runtime/blob/0363d81d2a355c2186df6efff6ee1742eaba3772/nmdc_runtime/api/endpoints/nmdcschema.py#L128) currently splits the string at each comma (`,`) and interprets the resulting substrings, as-is, as field names. 

For example, if the projection value is `id, ecosystem_type,myfield`, it'll split that into `id` and ` ecosystem_type` (where ` ecosystem_type` starts with a space) and `myfield`. Then, when it uses those to tell Mongo which fields it wants Mongo to include in the query result, Mongo sees ` ecosystem_type` (with the space) and `ecosystem_type` (no space) differently.

I think you're right that that [`get_pairs`](https://github.com/microbiomedata/nmdc-runtime/blob/0363d81d2a355c2186df6efff6ee1742eaba3772/nmdc_runtime/api/endpoints/util.py#L116) function could be used to make it so the projection value in the request can contain surrounding spaces. 👍 

CC: @dwinston

dwinston:
great work, @brynnz22 and @eecavanna. I just pushed https://github.com/microbiomedata/nmdc-runtime/commit/1a877d91b41851c9948cc95000b76209584e1a55 , which hopefully addresses both issues.

eecavanna:
> great work, @brynnz22 and @eecavanna. I just pushed [1a877d9](https://github.com/microbiomedata/nmdc-runtime/commit/1a877d91b41851c9948cc95000b76209584e1a55) , which hopefully addresses both issues.

Thanks @dwinston. That commit looks good to me. 👍

PR #292, opened 2023-08-22
fix: check user-supplied filter for processability

### LINKED ISSUE #290: The `/objects` endpoint responds with Internal Server Error for some `filter` values
Normal: The following request results in an HTTP 200 response. 🟢 
```clean
curl -X 'GET' \
  'https://api.microbiomedata.org/objects?max_page_size=20' \
  -H 'accept: application/json'
```
> This request does not include a `filter` parameter.

Normal: The following request results in an HTTP 200 response. 🟢 
```clean
curl -X 'GET' \
  'https://api.microbiomedata.org/objects?filter=%7B%22ecosystem_type%22%3A%20%22Freshwater%22%7D&max_page_size=20' \
  -H 'accept: application/json'
```
> This request has a `filter` parameter value of `{"ecosystem_type": "Freshwater"}`.

Problem: The following request results in an HTTP 500 response **instead of an HTTP 4xx response.** 🔴 
```clean
curl -X 'GET' \
  'https://api.microbiomedata.org/objects?filter=something&max_page_size=20' \
  -H 'accept: application/json'
```
> This request has a `filter` parameter value of `something`.

Steps to reproduce:

1. Go to https://api.microbiomedata.org/docs#/objects/list_objects_objects_get
2. Click the "Try it out" button
3. Populate the `filter` parameter with `something`
4. Click the "Execute" button
5. See the response is an HTTP 500 response

Screenshot:

image

dwinston:
closes #290

### DIFF
--- a/nmdc_runtime/api/endpoints/util.py
+++ b/nmdc_runtime/api/endpoints/util.py
@@ -59,13 +59,14 @@ HOSTNAME_EXTERNAL = BASE_URL_EXTERNAL.split("://", 1)[-1]
 
 
 def check_filter(filter_: str):
+    filter_ = filter_.strip()
     if not filter_.startswith("{") or not filter_.endswith("}"):
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-            detail=f"Given `filter` is not valid JSON object, which must start with '{{' and end with '}}'.",
+            detail=f"The given `filter` is not a valid JSON object, which must start with '{{' and end with '}}'.",
         )
     try:
-        d = json_util.loads(filter_)
+        json_util.loads(filter_)
     except JSONDecodeError as e:
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
--- a/nmdc_runtime/api/endpoints/util.py
+++ b/nmdc_runtime/api/endpoints/util.py
@@ -3,6 +3,7 @@ import os
 import re
 import tempfile
 from functools import lru_cache
+from json import JSONDecodeError
 from pathlib import Path
 from time import time_ns
 from typing import List, Optional, Set, Tuple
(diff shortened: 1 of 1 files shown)


nmdc_runtime/api/endpoints/util.py
```diff
@@ -57,9 +58,25 @@
 HOSTNAME_EXTERNAL = BASE_URL_EXTERNAL.split("://", 1)[-1]
 
 
+def check_filter(filter_: str):
+    if not filter_.startswith("{") or not filter_.endswith("}"):
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail=f"Given `filter` is not valid JSON object, which must start with '{{' and end with '}}'.",
```
eecavanna, line 65: >Given `filter` is not valid JSON object, ...

I started reading this as a dependent clause; like, "Given that `filter` is not a valid JSON object, ..."

Proposal:

The given `filter` is not a valid JSON object, ...

nmdc_runtime/api/endpoints/util.py
```diff
@@ -57,9 +58,25 @@
 HOSTNAME_EXTERNAL = BASE_URL_EXTERNAL.split("://", 1)[-1]
 
 
+def check_filter(filter_: str):
+    if not filter_.startswith("{") or not filter_.endswith("}"):
```
eecavanna, line 62: I don't know whether you want to allow the `filter` value to begin or end with whitespace. If so, I'd suggest:

```diff
- if not filter_.startswith("{") or not filter_.endswith("}"):
+ if not filter_.lstrip().startswith("{") or not filter_.rstrip().endswith("}"):
```

Docs:
- `lstrip()` - https://docs.python.org/3/library/stdtypes.html#str.lstrip

eecavanna (commented):
Thanks for adding this check, @dwinston! I left a couple comments. Neither one is a blocker from my perspective (in both cases, I consider it your call).

nmdc_runtime/api/endpoints/util.py
dwinston: cool. I added `filter_ = filter_.strip()` at the top of `check_filter` because I think it's fine to tolerate extra whitespace on the ends for user input.

nmdc_runtime/api/endpoints/util.py
dwinston: great. done.

PR #295, opened 2023-08-30
update `dev.txt` and `main.txt` based on NEON soil ingest pipeline requirements

sujaypatil96:
Swagger interface for API docs aren't being built properly because newly added `requests-cache` library is not in the environment.

### DIFF
--- a/nmdc_runtime/site/ops.py
+++ b/nmdc_runtime/site/ops.py
@@ -57,7 +57,7 @@ from nmdc_runtime.site.resources import (
     GoldApiClient,
     RuntimeApiSiteClient,
     RuntimeApiUserClient,
-    NeonApiClient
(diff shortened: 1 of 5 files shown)


nmdc_runtime/site/ops.py
```diff
@@ -589,23 +589,23 @@ def get_gold_study_pipeline_inputs(context: OpExecutionContext) -> str:
 
 @op(required_resource_keys={"gold_api_client"})
 def gold_biosamples_by_study(
-        context: OpExecutionContext, study_id: str
+    context: OpExecutionContext, study_id: str
```
dwinston, line 592: ran `black` for style formatting

dwinston (approved):

PR #296, opened 2023-08-31
Add instructions for setting `mongoKeyFile` permissions

eecavanna:
I added a new step to the instructions in the README:

image

I also "filled in" some of the existing steps with a little more information.


dwinston (approved):
thanks!

PR #303, opened 2023-09-06
397 add gold metadata to neon soil samples

mbthornton-lbl:
DRAFT

Draft PR with initial classes and tests for changesheet generators

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,26 +79,43 @@ database, as a hub for data management and workflow automation.
 
 ## Local Development
 
-Ensure Docker (and Docker Compose) are installed.
+Ensure Docker (and Docker Compose) are installed; and the Docker engine is running.
 
-Ensure you have a .env file for the docker services to source from. You may copy `.env.example` to
+```shell
+docker --version
+docker compose version
+docker info
+```
+
+Ensure the permissions of `./mongoKeyFile` are such that only the file's owner can read or write the file.
+
+```shell
+chmod 600 ./mongoKeyFile
+```
+
+Ensure you have a `.env` file for the Docker services to source from. You may copy `.env.example` to
 `.env` (which is gitignore'd) to get started.
 
-```
-# To load env in your shell session
-# export $(grep -v '^#' .env | xargs)
+```shell
+cp .env.example .env
 ```
 
-If you are connecting to resources that require a ssh tunnel, for example a MongoDB that is only
-accessible on the NERSC network,
+Create environment variables in your shell session, based upon the contents of the `.env` file.
 
+```shell
+export $(grep -v '^#' .env | xargs)
 ```
+
+If you are connecting to resources that require an SSH tunnel—for example, a MongoDB server that is only accessible on the NERSC network—set up the SSH tunnel.
+
+The following command could be useful to you, either directly or as a template (see `Makefile`).
+
+```shell
 make nersc-ssh-tunnel
 ```
 
-could be useful for you, directly or as a template.
+Finally, spin up the Docker Compose stack.
 
-Finally,
 ```bash
 make up-dev
 ```
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -74,7 +74,7 @@ services:
       - .:/code
 
   mongo:
-    image: bitnami/mongodb:6.0.4
+    image: mongo:6.0.4
     container_name: mongo
     ports:
       - "27018:27017"
@@ -84,11 +84,8 @@ services:
       - ./tests/mongorestore-nmdc-testdb.sh:/mongorestore-nmdc-testdb.sh:ro
     restart: unless-stopped
     environment:
-      MONGODB_ROOT_USER: admin
-      MONGODB_ROOT_PASSWORD: root
-      MONGODB_REPLICA_SET_MODE: primary
-      MONGODB_ADVERTISED_HOSTNAME: mongodb-primary
-      MONGODB_REPLICA_SET_KEY: replicasetkey123
+      MONGO_INITDB_ROOT_USERNAME: admin
+      MONGO_INITDB_ROOT_PASSWORD: root
 
   test:
     build:
--- a/nmdc_runtime/site/changesheets/changesheet_generator.py
+++ b/nmdc_runtime/site/changesheets/changesheet_generator.py
@@ -60,7 +60,7 @@ class BaseChangesheetGenerator(ChangesheetGenerator):
     Base class for changesheet generators
     """
 
-    def __init__(self, name) -> None:
+    def __init__(self, name: str) -> None:
         self.name = name
         self.changesheet = Changesheet()
         self.output_filename_root = f"{self.name}-{time.strftime('%Y%m%d-%H%M%S')}"
--- a/nmdc_runtime/site/changesheets/changesheet_generator.py
+++ b/nmdc_runtime/site/changesheets/changesheet_generator.py
@@ -1,0 +1,94 @@
+# nmdc_runtime/site/changesheets/changesheet_generator.py
+"""
+changesheet_generator.py: Provides classes to generate and validate changesheets for NMDC database objects.
+"""
+from abc import ABC, abstractmethod
+from dataclasses import dataclass, field
+import time
+from typing import ClassVar, Dict, Any
+
+JOSN_OBJECT = Dict[str, Any]  # TODO: de-duplicate this with the one in translator.py
+
+
+@dataclass
+class ChangesheetLineItem:
+    """ Dataclass representing a line-item in a changesheet"""
+    id: str
+    action: str
+    attribute: str
+    value: str
+
+    @property
+    def line(self) -> str:
+        return f"{self.id}\t{self.action}\t{self.attribute}\t{self.value}"
+
+
+@dataclass
+class Changesheet:
+    """ Dataclass representing a changesheet being generated """
+    header: ClassVar[str] = "id\taction\tattribute\tvalue"
+    line_items: list[ChangesheetLineItem] = field(default_factory=list)
+
+
+class ChangesheetGenerator(ABC):
+    """
+    Abstract base class for changesheet generators
+    """
+
+    def __int__(self) -> None:
+        pass
+
+    @abstractmethod
+    def validate_changesheet(self) -> bool:
+        """
+        Validate a changesheet against the NMDC runtime validation API
+        :return: bool
+        """
+        pass
+
+    @abstractmethod
+    def write_changesheet(self) -> None:
+        """
+        Write a changesheet to a file in .tsv format
+        :return: None
+        """
+        pass
+
+
+class BaseChangesheetGenerator(ChangesheetGenerator):
+    """
+    Base class for changesheet generators
+    """
+
+    def __init__(self, name) -> None:
+        self.name = name
+        self.changesheet = Changesheet()
+        self.output_filename_root = f"{self.name}-{time.strftime('%Y%m%d-%H%M%S')}"
+
+    def add_changesheet_line_item(self, line_item: ChangesheetLineItem) -> None:
+        """
+        Add a line item to the changesheet
+        :param line_item: ChangesheetLineItem
+        :return: None
+        """
+        self.changesheet.line_items.append(line_item)
+
+    def validate_changesheet(self) -> bool:
+        """
+        Validate a changesheet against the NMDC runtime validation API
+        :return: bool
+        """
+        return NotImplemented
+
+    def write_changesheet(self, output_filename=None) -> None:
+        """
+        Write a changesheet to a file in .tsv format
+        :return: None
+        """
+        if not output_filename:
+            output_filename = f"{self.output_filename_root}.tsv"
+
+        with open(output_filename, "w") as f:
+            f.write(self.changesheet.header)
+            for line_item in self.changesheet.line_items:
+                f.write(line_item.line)
--- a/nmdc_runtime/site/changesheets/gold_changesheet_generator.py
+++ b/nmdc_runtime/site/changesheets/gold_changesheet_generator.py
@@ -12,9 +12,19 @@ class BaseGoldBiosampleChangesheetGenerator(BaseChangesheetGenerator):
     Class for generating changesheets for GOLD-derived metadata
     """
 
-    def __init__(self, gold_biosamples: JOSN_OBJECT) -> None:
-        super().__init__()
+    def __init__(self, name: str, gold_biosamples: list[JOSN_OBJECT]) -> None:
+        super().__init__(name)
         self.gold_biosamples = gold_biosamples
+        
+        self.gold_biosample_names= [biosample["biosampleName"].split()[-1] for biosample in gold_biosamples]
+        
+    def get_biosample_name(self, biosample: JOSN_OBJECT) -> str:
+        """
+        Get the name of a biosample
+        :param biosample: a biosample object
+        :return: str
+        """
+        return biosample["biosampleName"].split()[-1]
 
 
 class Issue397ChangesheetGenerator(BaseGoldBiosampleChangesheetGenerator):
@@ -23,5 +33,5 @@ class Issue397ChangesheetGenerator(BaseGoldBiosampleChangesheetGenerator):
     """
     issue_link = "https://github.com/microbiomedata/issues/issues/397"
 
-    def __init__(self, gold_biosamples: JOSN_OBJECT) -> None:
-        super().__init__(gold_biosamples)
+    def __init__(self, name: str, gold_biosamples: list[JOSN_OBJECT]) -> None:
+        super().__init__(name, gold_biosamples)
--- a/nmdc_runtime/site/changesheets/gold_changesheet_generator.py
+++ b/nmdc_runtime/site/changesheets/gold_changesheet_generator.py
@@ -1,0 +1,27 @@
+# nmdc_runtime/site/change_sheets/gold_changesheet_generator.py
+"""
+gold_changesheet_generator.py: Provides classes to generate and validate changesheets for NMDC database objects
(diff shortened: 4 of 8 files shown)


docs/nb/get_data.ipynb
mbthornton-lbl, line 1: I did not intend to / don't know how I managed to make changes to this file :-|

nmdc_runtime/test.Dockerfile
```diff
@@ -28,10 +28,10 @@ RUN pip install --no-cache-dir -r /code/requirements.txt
 COPY . /code
 RUN pip install --no-cache-dir --editable .
 
-# Ensure wait-for-it
-RUN chmod +x wait-for-it.sh
+## Ensure wait-for-it
+#RUN chmod +x wait-for-it.sh
```
dwinston, line 32: why?

neon_cache.sqlite
eecavanna, line 1: I am wondering whether this `.sqlite` file was committed by accident. It is listed in the `.gitignore` file, but it might have been committed before it was listed there.

nmdc_runtime/site/changesheets/base.py
```diff
@@ -0,0 +1,80 @@
...
+import requests
+from typing import Any, ClassVar, Dict
+
+from nmdc_runtime.site.resources import RuntimeApiUserClient
+
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s %(levelname)s %(" "message)s"
+)
+
+JSON_OBJECT = Dict[str, Any]
```
eecavanna, line 19: I like this type alias. I find it easier to mentally process than the `Dict[str, Any]` that it aliases.

I assumed the variable defined on the line below it was also a type alias (due to its proximity and similar formatting), but I don't think it is. I suggest adding `: TypeAlias` to the first one; i.e.

```diff
- from typing import Any, ClassVar, Dict
+ from typing import Any, ClassVar, Dict, TypeAlias

(...)

- JSON_OBJECT = Dict[str, Any]
+ JSON_OBJECT: TypeAlias = Dict[str, Any]
```
> Docs: https://docs.python.org/3/library/typing.html#typing.TypeAlias

nmdc_runtime/site/changesheets/base.py
```diff
@@ -0,0 +1,80 @@
...
+        self.output_filepath: Path = self.output_dir.joinpath(self.output_filename)
+
+    def validate_changesheet(self, client: RuntimeApiUserClient) -> bool:
+        """
+        Validate the changesheet
+        :return: None
+        """
+        logging.info(f"Validating changesheet {self.output_filepath}")
+        resp = requests.post(
+            "https://api.microbiomedata.org/metadata/changesheets:validate",
```
eecavanna, line 66: I'm wondering whether there may be a situation where the person running this code want to use a different base URL (e.g. the dev environment versus the prod environment; or a different environment entirely). I suggest making the base URL be dynamic (maybe basing it upon something in an environment variable, or at least a constant at the top of the module).

nmdc_runtime/site/changesheets/scripts/missing_neon_soils_ecosystem_data.py
```diff
@@ -0,0 +1,299 @@
...
+            value=gold_project_id,
+        )
+
+
+@click.command()
+@click.option("--study_id", default=GOLD_NEON_SOIL_STUDY_ID, help="GOLD study ID")
+@click.option(
+    "--apply_changes", is_flag=True, default=False, help=("Apply the" "changes")
+)
+def generate_changesheet(study_id, apply_changes):
```
eecavanna, line 213: I learned a couple days ago that `click` will display the function's docstring in the CLI command's help message. I recommend adding a docstring to this function, that describes what the CLI command does.

For example:
```py
def generate_changesheet(study_id, apply_changes):
    """Generates an X by doing Y with Z...."""
```

nmdc_runtime/site/changesheets/scripts/missing_neon_soils_ecosystem_data.py
```diff
@@ -0,0 +1,299 @@
...
+            action="insert item",
+            attribute="alternative_identifiers",
+            value=gold_project_id,
+        )
+
+
+@click.command()
+@click.option("--study_id", default=GOLD_NEON_SOIL_STUDY_ID, help="GOLD study ID")
+@click.option(
+    "--apply_changes", is_flag=True, default=False, help=("Apply the" "changes")
```
eecavanna, line 211: The `help=("Apply the" "changes")` value appears to me to be a copy/paste error or the result of a code formatter quirk.

nmdc_runtime/site/changesheets/scripts/missing_neon_soils_ecosystem_data.py
```diff
@@ -0,0 +1,299 @@
...
+    logging.debug(
+        f"Read {lines_read} lines from OmicsProcessing-to-catted-Biosamples.tsv"
+    )
+    logging.info(
+        f"Skipped {lines_skipped} lines from OmicsProcessing-to-catted-Biosamples.tsv"
+    )
+    return omics_processing_to_biosamples
+
+
+def goldbs_to_nmdcbs_and_omprc(
```
eecavanna, line 71: Since this function's name contains some abbreviations that I'm not familiar with (took me a few seconds to translate them into "gold biosamples to NMDC biosamples and omics processings"), I'd prefer the function name be more "spelled out" or at least have a docstring that explains its functionality.

docs/nb/get_data.ipynb
eecavanna: You can (effectively) remove these changes from the PR branch by adding a new commit that contains the original, unedited file.

neon_cache.sqlite
mbthornton-lbl: By accident for sure -

nmdc_runtime/site/changesheets/scripts/missing_neon_soils_ecosystem_data.py
mbthornton-lbl: why not both?

nmdc_runtime/test.Dockerfile
mbthornton-lbl: no good reason - removed

nmdc_runtime/site/changesheets/base.py
mbthornton-lbl: updated validate_changesheet to require a base_url argument, taken from .env by the caller

nmdc_runtime/site/changesheets/scripts/missing_neon_soils_ecosystem_data.py
mbthornton-lbl: done

nmdc_runtime/site/changesheets/scripts/missing_neon_soils_ecosystem_data.py
mbthornton-lbl: Done

docs/nb/get_data.ipynb
mbthornton-lbl: Done

neon_cache.sqlite
eecavanna, line 1: This `neon_cache.sqlite` file still exists, as of the latest commit on this branch.

eecavanna: You can delete it by issuing the following commands on your computer:
```shell
# Assuming you have this branch checked out and are in the root folder of the repo:
$ git rm neon_cache.sqlite
$ git commit -m "Delete neon_cache.sqlite"

# Then, push that new commit up to GitHub.
```

demo/metadata_migration/notebooks/helpers.py
eecavanna, line 1: Other reviewers can ignore the changes in this file, which all seem to be `black` formatting (i.e. no functional changes).

neon_cache.sqlite
mbthornton-lbl: Done

nmdc_runtime/site/changesheets/base.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+
+    id: str
+    action: str
+    attribute: str
+    value: str
+
+    @property
+    def line(self) -> str:
+        return f"{self.id}\t{self.action}\t{self.attribute}\t{self.value}"
```
eecavanna, line 36: I am wondering whether this accounts for the possibility that the "cell" values, themselves, contains tab characters. 

I see this on [the Wikipedia page about the TSV format](https://en.wikipedia.org/wiki/Tab-separated_values#Character_escaping):

> The [IANA](https://en.wikipedia.org/wiki/Internet_Assigned_Numbers_Authority) [media type](https://en.wikipedia.org/wiki/Media_type) standard for TSV achieves simplicity by simply disallowing tabs within fields.

I don't know whether the person who wrote that meant that there is no way to represent a tab-containing value within a TSV file. 🤷 

Caveat: I haven't looked at the code that calls this function / uses this class yet.

neon_cache.sqlite
eecavanna: Confirmed. Thanks!

nmdc_runtime/site/changesheets/base.py
eecavanna: Based on my own _guess_ at the requirements for this: I think `raise`-ing an `ValueError` having an error message like "We do not allow values to contain tabs or newlines" would be an adequate way to handle that situation.

docs/nb/get_data.ipynb
mbthornton-lbl, line 1: Format change from blackening

nmdc_runtime/site/changesheets/base.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@dataclass
+class Changesheet:
+    """
+    A changesheet
+    """
+
+    name: str
+    line_items: list = None
```
eecavanna, line 46: I think `None` is not a valid value of type `list`.

You could either change the value to `[]`, or change the type to either `list | None` ([docs](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#x-y-syntax-for-unions)) or `Optional[list]` ([docs](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#optional-types-and-the-none-type)). `Optional` comes from `typing`; as in: `from typing import Optional`

nmdc_runtime/site/changesheets/base.py
```diff
@@ -0,0 +1,81 @@
...
+    header: ClassVar[str] = "id\taction\tattribute\tvalue"
+    output_dir: Path = None
```
eecavanna, line 48: Consider changing to `Optional[Path]`

nmdc_runtime/site/changesheets/scripts/missing_neon_soils_ecosystem_data.py
```diff
@@ -0,0 +1,338 @@
...
+def _check_gold_biosample_identifiers(goldbs, nmdcbs) -> List[ChangesheetLineItem]:
+    changesheet_line_items = []
+    goldbs_id = normalize_gold_id(goldbs["biosampleGoldId"])
+    if not goldbs_id in nmdcbs["gold_biosample_identifiers"]:
+        changesheet_line_items.append(
+            ChangesheetLineItem(
+                id=nmdcbs["id"],
+                action="insert",
+                attribute="gold_biosample_identifiers",
+                value=goldbs_id + "|",
```
eecavanna, line 193: This reminds me: an existing value could contain a pipe character (I mean a string, in general, could contain one—not to say our schema would permit it to for a given field). I don't know whether our changesheet functionality provides a way to escape a pipe character. 🤷

nmdc_runtime/site/changesheets/scripts/missing_neon_soils_ecosystem_data.py
```diff
@@ -0,0 +1,338 @@
...
+            action="insert",
+            attribute="gold_sequencing_project_identifiers",
+            value=gold_project_id + "|",
+        )
+
+
+@click.command()
+@click.option("--study_id", default=GOLD_NEON_SOIL_STUDY_ID, help="GOLD study ID")
+@click.option(
+    "--apply_changes", is_flag=True, default=False, help=("Apply the changes")
```
eecavanna, line 221: I recommend adding the "Not implemented!" caveat to the option's own `help` string. I almost missed it in the main function's docstring.

nmdc_runtime/site/changesheets/scripts/missing_neon_soils_ecosystem_data.py
```diff
@@ -0,0 +1,338 @@
...
+            nmdc_biosamples,
+            omprc_records,
+        ) = gold_biosample_to_nmdc_biosamples_and_omics_processing_records(
+            runtime_api_user_client, omprc_to_bs_map, goldbs
+        )
+        if not nmdc_biosamples:
+            logging.warning(
+                f"no corresponding NMDC biosamples found for {goldbs['biosampleGoldId']}..."
+            )
+            unfindable_count += 1
```
eecavanna, line 291: I think this is always the length of `unfindable_goldbs_id` (one line later) and could be computed on-demand (in the `logging.info` statement below).

tests/test_normalization/test_gold.py
eecavanna, line 1: Thanks for including tests!

nmdc_runtime/site/normalization/gold.py
```diff
@@ -0,0 +1,43 @@
...
+
+def get_gold_biosample_name_suffix(biosample_name: str) -> str:
+    """
+    Get the suffix for the name of a GOLD biosample - the last word in the biosampleName attribute
+    e.g. "biosampleName": "Terrestrial soil microbial communities from
+    Disney Wilderness Preserve, Southeast, FL, USA - DSNY_016-M-37-14-20140409-GEN-DNA1",
+
+    Suffix = DSNY_016-M-37-14-20140409-GEN-DNA1
+
+    :param biosample: a list of JSON objects representing GOLD biosamples
```
eecavanna, line 24: I think this `:param` name and description is obsolete.

Suggestion:
```diff
- :param biosample: a list of JSON objects representing GOLD biosamples
+ :param biosample_name: name of a biosample
```

nmdc_runtime/site/normalization/gold.py
```diff
@@ -0,0 +1,43 @@
...
+    :return: str
+    """
+    return biosample_name.split()[-1]
+
+
+def normalize_gold_id(gold_id: str) -> str:
+    """
+    Normalize the given GOLD biosample ID to the form "GOLD:<gold_biosample_id>"
```
eecavanna, line 32: I think this description is not accurate (I think the output starts with lowercase `gold:`).

nmdc_runtime/site/normalization/gold.py
```diff
@@ -0,0 +1,43 @@
...
+    :param gold_id: str
+    :return: str
+    """
+    if gold_id.startswith("gold:"):
+        return gold_id
+    elif gold_id.startswith("GOLD:") or gold_id.startswith("Gold:"):
```
eecavanna, line 38: In case you also want to catch `gOld`, `goLd`, `GoLD`, etc., you can do this:

```diff
- elif gold_id.startswith("GOLD:") or gold_id.startswith("Gold:"):
+ elif gold_id.upper().startswith("GOLD:"):
```
 - Docs: https://docs.python.org/3/library/stdtypes.html#str.upper

You can assume it is not `gold:` since the previous condition would have caught that.

eecavanna:
Looking good, @mbthornton-lbl! Thanks for adding all the docs and type information. Aside from what I wrote in the review comments, I am comfortable with this being merged in. I will defer to @dwinston regarding actually merging it because I haven't worked with the `site` folder before and am not sure what he has in mind regarding creating new folders in there (e.g. the `changesheets` folder). He may want this code to reside somewhere else.

nmdc_runtime/site/changesheets/base.py
mbthornton-lbl: What about stripping / de-tabbing `value`?  I can't think of a legitimate case where there SHOULD be a tab character in value

nmdc_runtime/site/changesheets/base.py
mbthornton-lbl: It's actually `line_items: list = field(default_factory=list)`

nmdc_runtime/site/changesheets/scripts/missing_neon_soils_ecosystem_data.py
mbthornton-lbl: In this case we are only doing this with gold_biosample_identifiers, which have a defined format e.g. bs1123456 so I think this can stay as-is

nmdc_runtime/site/changesheets/scripts/missing_neon_soils_ecosystem_data.py
mbthornton-lbl: done

nmdc_runtime/site/normalization/gold.py
mbthornton-lbl: Done.

nmdc_runtime/site/normalization/gold.py
mbthornton-lbl: Updated.

nmdc_runtime/site/normalization/gold.py
mbthornton-lbl: Done.

dwinston (approved):
There's a lot of refactoring I'd like to see happen, but given that the functionality herein is limited to producing artifacts (changesheets) that are to be subsequently vetted and submitted separately, that is, nothing in this PR affects existing functionality / other processes, I see no reason to request further work on this. Thank you for isolating your significant changes to new modules.

PR #313, opened 2023-09-22
Convert API endpoint to a no-op and omit from Swagger UI

### LINKED ISSUE #312: API endpoint (`/sites/{site_id}/capabilities`) responds with HTTP 500 Internal Server Error
The `/sites/{site_id}/capabilities` endpoint always responds with an HTTP 500 Internal Server Error.

@shreddd and I saw that the function that handles the request accepts and returns a string, but its type information says it will return a list; and I think that string-to-list serialization is failing, causing the server-side error.

https://github.com/microbiomedata/nmdc-runtime/blob/712505ed77ef5912c609a02a65944e1bfb3a23a6/nmdc_runtime/api/endpoints/sites.py#L101-L103

#### Response on client:

image

#### Error on server:

image

@shreddd suggested converting the function into a no-op/TODO (like some of the functions around it; example below)—given that its core functionality isn't implemented yet—so it doesn't trigger a server-side error.

https://github.com/microbiomedata/nmdc-runtime/blob/712505ed77ef5912c609a02a65944e1bfb3a23a6/nmdc_runtime/api/endpoints/sites.py#L95-L98

eecavanna:
Fixes #312

### DIFF
--- a/nmdc_runtime/api/endpoints/sites.py
+++ b/nmdc_runtime/api/endpoints/sites.py
@@ -98,9 +98,10 @@ def replace_site():
     pass
 
 
-@router.get("/sites/{site_id}/capabilities", response_model=List[Capability])
+@router.get("/sites/{site_id}/capabilities", include_in_schema=False)
 def list_site_capabilities(site_id: str):
-    return site_id
+    """Not yet implemented"""
+    pass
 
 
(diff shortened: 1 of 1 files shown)


dwinston (approved):
Sounds good. And it's cleaner to have zero "capabilities"-related endpoints unless and until the whole suite is implemented. And I realized the remaining endpoint was incorrectly designated.

PR #321, opened 2023-10-05
fix: refresh minter.requesters collection on site creation

### LINKED ISSUE #317: sites are cached by the minter at container startup
Sites are cached by the minter when the runtime container is started. When a new site is created and credentials are made and then a user tries to use the minter endpoint the code generates a 422 error with an error message of "Unknown requester $new_site" 

suggested resolution: the minter endpoint should refresh the list of sites at a regular interval

To reproduce:
Generate a new site with https://api.microbiomedata.org/docs#/sites/create_site_sites_post (ie minter_cache_test)
Generate credentials with https://api.microbiomedata.org/docs#/sites/generate_credentials_for_site_client_sites__site_id__generateCredentials_post
login w/client credentials 
use the minter endpoint to try to mint an id of any type w/o restarting rancher
https://api.microbiomedata.org/docs#/minter/mint_ids_pids_mint_post
Expected return error "Unknown requester minter_cache_test"

@eecavanna  did restart rancher yesterday and then I was able to mint an ID from the newly created site which is what @dwinston  had suggested to test the caching hypothesis. 

cc @brynnz22 @shreddd

dwinston:
The interface to modifying the sites collection is limited to a single endpoint, so I think it is fine to directly call a function to perform the operation.

I can envision moving to a more event-driven handler architecture once enough of these cases are present. I don't think it's worth the added complexity at this time.

fixes #317

### DIFF
--- a/nmdc_runtime/api/endpoints/sites.py
+++ b/nmdc_runtime/api/endpoints/sites.py
@@ -40,6 +40,7 @@ from nmdc_runtime.api.models.site import (
 )
 from nmdc_runtime.api.models.user import get_current_active_user, User
 from nmdc_runtime.api.models.util import ListResponse, ListRequest
+from nmdc_runtime.minter.bootstrap import refresh_minter_requesters_from_sites
 
 router = APIRouter()
 
@@ -56,6 +57,7 @@ def create_site(
             detail=f"site with supplied id {site.id} already exists",
         )
     mdb.sites.insert_one(site.dict())
+    refresh_minter_requesters_from_sites()
     rv = mdb.users.update_one(
         {"username": user.username},
         {"$addToSet": {"site_admin": site.id}},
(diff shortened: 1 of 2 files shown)


dwinston:
https://github.com/microbiomedata/nmdc-runtime/blob/a5338b176ca4c54f4d281a51fde51e7fb5b7ab5e/nmdc_runtime/minter/bootstrap.py#L19

the code in the bootstrap function from this line down needs to be re-run if the sites collection changes, so need to factor out this code into its own function and ensure it is called both on runtime bootstrap and on `POST /site`, the only other part of the codebase that dynamically (after boot) alters the sites collection.

eecavanna (approved):
I agree with what @dwinston said in the issue description: Given that only one endpoint modifies the sites collection, I think "manually" calling the refresh function in that endpoint is sufficient.

PR #323, opened 2023-10-10
fix: mongo limit wrt `distinct`

dwinston:
Avoid attempt to fetch large (>16mb) list of distinct IDs, a limitation of collection.distinct(). 
Use aggregation pipeline instead to compute on mongo server, using disk if necessary.

/cc @shreddd


eecavanna (approved):
LGTM!

Issue #315, opened 2023-09-25
Import Andrew Tritt's submission into Mongo


pkalita-lbl:
Andrew Tritt has said that his submission (https://data.microbiomedata.org/submission/80f5752b-e8d3-453f-9a47-f48210ed6ffc/samples) is ready to go. Montana has given her approval. This task is to:

- [x] Attempt the submission-into-mongo import in a dev/local environment
- [x] Make any necessary updates to the pipeline
- [ ] Run the pipeline in production environment

pkalita-lbl:
End of sprint update: the first two tasks are done. Next steps are to review the local import with Montana, commit changes to `nmdc-runtime`, and run the pipeline on production. Will complete next sprint.

pkalita-lbl:
Also @aclum just sent me information that should allow us to create `omics_processing` and `data_object` records. So I will look into also updating the submission-into-mongo pipeline code to accept a mapping file and build such objects out of the information therein.

pkalita-lbl:
Necessary code changes have been made to `nmdc-runtime`, but I want to keep this open until we actually run the pipeline on production. Running it is currently blocked by some Rancher secret issues that I'm still working to resolve. Moving to next sprint.

pkalita-lbl:
* Secrets issue resolved on Rancher
* Ran pipeline on production Dagster
* Ran data portal ingest on production
* We have a study page: https://data.microbiomedata.org/details/study/nmdc:sty-11-hdd4bf83
* Alerted Mark F and Michal that OmicsProcessing records are available for them to pick up and haven't heard any issues from them.

I think it's finally done.

Issue #301, opened 2023-09-02
Multiple `pytest` tests fail and testing hangs


eecavanna:
### Description

When I run `pytest` without the `-x` option (i.e. when I add the following* to `docker-compose.test.yml` and then run `$ make up-test` followed by `$ make test`), **multiple tests fail... and testing hangs**/freezes.

*here's what I add to `docker-compose.test.yml` so that `$ make test` causes `pytest` to run without the `-x` option (this overrides the `ENTRYPOINT` written in `test.Dockerfile`, which includes the `-x` option):

```diff
  test:
    build:
      context: .
      dockerfile: nmdc_runtime/test.Dockerfile
    container_name: test
+   entrypoint:
+     - "./wait-for-it.sh"
+     - "fastapi:8000"
+     - "--strict"
+     - "--timeout=300"
+     - "--"
+     - "pytest"
    env_file:
```

### Screenshot

Here's a screenshot of the output of `$ make test`, with the above modification to `docker-compose.test.yml` in place. This shows that multiple tests have failed. Also, testing has hung/froze at the point shown (in the sense that, after waiting for more than 10 minutes—specifically timed—there was no change in the output).

![image](https://github.com/microbiomedata/nmdc-runtime/assets/134325062/88e19f45-6a7a-4b63-b68c-02dfb3f8ff7d)

### Environment:
- Docker host: Apple M1 Pro running macOS Ventura 13.4.1 (c)
- Docker Desktop: 4.21.1 (114176) 
- Docker Engine: 24.0.2
- Docker Compose: v2.19.1
- Git commit f812f7bc303ba840c93b4d7dd7dfef4665e1e3d8 (plus the change to `docker-compose.test.yml` shown above)

eecavanna:
I pressed `^C` (once or twice) to terminate the testing. 

![image](https://github.com/microbiomedata/nmdc-runtime/assets/134325062/f3f4e1d3-3a60-491b-aee9-708b19f16d74)

Once I did that, the `==== FAILURES ====` section appeared. It showed information about each of the 3 failures that had occurred up to that point. I will post about each one below.

eecavanna:
### 1. test_nmdc_jsonschema_using_new_id_scheme

```py
________________________________________________________________________________________________________________________ test_nmdc_jsonschema_using_new_id_scheme ________________________________________________________________________________________________________________________

    def test_nmdc_jsonschema_using_new_id_scheme():
[...]

PR #329, opened 2023-10-18
Updates to submission portal jobs to handle sequencing data

pkalita-lbl:
Re: #315 (will close once we actually run the job on production)

We previously had jobs to fetch a submission portal entry, translate it into `Study` and `Biosample` objects, validate, and submit to Mongo. It turns out Andrew Tritt's submission (and possibly future ones??) came with sequencing data. The submission portal doesn't have the capability to accept that information, so it was provided in a CSV file mapping sample IDs (that is, the ID provided in the submission portal, not an NMDC minted ID) to FASTA files.

These changes:
* Update the submission portal jobs to accept two CSV files where the columns correspond to `OmicsProcessing` and `DataObject` slots, with one extra column to join on sample IDs. If URLs to such files are provided in the job config, each row will be expanded into the corresponding objects by the `SubmissionPortalTranslator` class.
* The `SubmissionPortalTranslator` class and corresponding tests were also a little behind-the-times on some recent schema changes (DOI stuff, ID patterns, etc). That has been resolved, and the test now does a schema validation step to catch these things earlier (especially if we, ahem, ran tests against PRs in CI)
* I cut out a bunch of sample data from the existing test case in `tests/test_data/test_submission_portal_translator_data.yaml` -- it was redundant as far as testing is concerned and it made test failure diffs hard to read. I added a second test case in that file to cover the case of providing `OmicsProcessing` and `DataObject` mapping data.


nmdc_runtime/site/translation/submission_portal_translator.py
```diff
@@ -51,19 +63,17 @@ def _get_pi(
...
-    def _get_doi(
-        self, metadata_submission: JSON_OBJECT
-    ) -> Union[nmdc.AttributeValue, None]:
-        """Construct an nmdc:AttributeValue object using information from the context form data
+    def _get_doi(self, metadata_submission: JSON_OBJECT) -> Union[List[str], None]:
+        """Get DOI information from the context form data
 
         :param metadata_submission: submission portal entry
-        :return: nmdc:AttributeValue
+        :return: string or None
```
eecavanna, line 70: I think this is inaccurate.

```diff
-         :return: string or None
+         :return: list or None
```

nmdc_runtime/site/translation/submission_portal_translator.py
```diff
@@ -364,6 +374,38 @@ def _transform_value_for_slot(self, value: Any, slot: SlotDefinition):
 
         return transformed_value
 
+    def _transform_dict_for_class(self, raw_values: dict, class_name: str) -> dict:
+        """Transform a dict of values according to class slots.
+
+        raw_values is a dict where the keys are slot names and the values are plain strings.
+        Each of the items in this dict will be transformed by the _transform_value_for_slot
+        method. If the slot is multivalued each individual value will be transformed. If the
+        slot is multivalued and the value is a string it will be split in pipe characters
```
eecavanna, line 383: Suggestion:

```diff
-        ... split in pipe characters
+        ... split at pipe characters
```

nmdc_runtime/site/translation/submission_portal_translator.py
```diff
@@ -364,6 +374,38 @@ def _transform_value_for_slot(self, value: Any, slot: SlotDefinition):
...
+        transformed_values = {}
+        for column, value in raw_values.items():
+            if column not in slot_names:
+                logging.warning(f"No slot '{column}' on class '{class_name}'")
+                continue
+
+            slot_definition = self.schema_view.induced_slot(column, class_name)
+            if slot_definition.multivalued:
+                if isinstance(value, str):
+                    value = [v.strip() for v in value.split("|")]
```
eecavanna, line 396: I think I'd find this code block easier to read if the returned list was named `items` or `subvalues` (something plural).

tests/test_data/test_submission_portal_translator.py
```diff
@@ -234,18 +240,37 @@ def test_get_from():
...
+    )
+
+    mongo_db = get_mongo_test_db()
     with open(
         Path(__file__).parent / "test_submission_portal_translator_data.yaml"
     ) as f:
         test_datasets = yaml.safe_load_all(f)
 
         for test_data in test_datasets:
+            random.seed(0)
```
eecavanna, line 266: I'm not seeing where randomness comes into play in this test. Maybe one of the called functions relies on it.

nmdc_runtime/site/translation/submission_portal_translator.py
```diff
@@ -444,4 +472,87 @@ def get_database(self) -> nmdc.Database:
...
+                # Transform the raw row data according to the OmicsProcessing class's slots, and
+                # generate an instance. A few key slots do not come from the mapping file, but
+                # instead are defined here.
+                omics_processing_slots = {
+                    "id": self._id_minter("nmdc:OmicsProcessing", 1)[0],
+                    "has_input": [nmdc_biosample_id],
+                    "has_output": [],
+                    "part_of": nmdc_study_id,
+                    "add_date": today,
+                    "mod_date": today,
```
eecavanna, line 525: I don't know whether the person who defined these slots expects the date to be in a particular timezone. I see, above, that the value is defined as `today = datetime.now().strftime("%Y-%m-%d")`. I think `datetime.now()` returns the local time with respect to the server (as opposed to, for example, the UTC time).

nmdc_runtime/site/translation/submission_portal_translator.py
```diff
@@ -444,4 +472,87 @@ def get_database(self) -> nmdc.Database:
...
+                    self._transform_dict_for_class(
+                        omics_processing_row, "OmicsProcessing"
+                    )
+                )
+                omics_processing = nmdc.OmicsProcessing(**omics_processing_slots)
+
+                for data_object_row in data_objects_by_sample_data_id.get(
+                    sample_data_id, []
+                ):
+                    # For each row in the DataObject mapping file that correspond to the sample ID,
```
eecavanna, line 538: Typo:

```diff
- ... that correspond to...
+ ... that corresponds to...
```

eecavanna (approved):
Thanks for updating the tests! I wonder how it'll affect https://github.com/microbiomedata/nmdc-runtime/issues/301.

Looks good to me (I reviewed it in terms of "do I understand what this does in isolation"), with the caveat that I don't have experience with the submission portal or this part of the codebase (i.e. "do I understand what this does to the big picture").

nmdc_runtime/site/ops.py
```diff
@@ -773,3 +806,15 @@ def id_minter(*args, **kwargs):
...
+@op
+def get_csv_file_from_url(url: str) -> List[Dict]:
+    if not url:
+        return []
+
+    response = requests.get(url)
+    response.raise_for_status()
+
+    reader = csv.DictReader(response.text.splitlines())
+    return [valfilter(identity, row) for row in reader]
```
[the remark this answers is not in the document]
pkalita-lbl: You've understood it correctly. It's a very terse way of saying, please turn this:

```python
{
  "a": "",
  "b": "hello",
  "c": "",
  "d": "world"
}
```

into this:

```python
{
  "b": "hello",
  "d": "world"
}
```

We need that because downstream schema-based classes don't have the same empty string / null equivalency that CSV does. So we need to get those empty strings out of there before passing these dicts to those classes.

Regardless, your point stands that a little comment wouldn't hurt.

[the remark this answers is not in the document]
eecavanna: Thanks for confirming and rephrasing. 

For the code comment, I was originally thinking of it being on the same line as (or the line above) the statement in question; but I also think adding an overall function docstring and mentioning it there instead would be sufficient.

tests/test_data/test_submission_portal_translator.py
pkalita-lbl: The `test_minter` [fixture](https://github.com/microbiomedata/nmdc-runtime/blob/29c23ffb54ca2ce1da23bb43d15e658478ceee44/tests/test_data/conftest.py#L10) generates random IDs. Resetting the seed keeps the generated IDs stable between test runs.

nmdc_runtime/site/translation/submission_portal_translator.py
pkalita-lbl: That's a fair question. As far as I can tell from the existing data there is no particular standard. The schema doesn't [explicitly state](https://microbiomedata.github.io/nmdc-schema/add_date/) anything either. This might be something for further discussion later.

tests/test_data/test_submission_portal_translator.py
eecavanna: Ah, OK. That makes sense. Thanks!

nmdc_runtime/site/translation/submission_portal_translator.py
eecavanna: OK, cool. Thanks for linking to the slot documentation—I don't see any info about timezones there either. I'll file a ticket in `nmdc-schema` as a starting point.

nmdc_runtime/site/ops.py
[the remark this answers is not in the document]
pkalita-lbl: Comments added.

eecavanna (approved):
You addressed all my concerns. Thanks!

PR #338, opened 2023-10-31
fix: pin to fastapi>=0.104.1 to pin swagger ui

### LINKED ISSUE #337: swagger API pins CPUs and freezes
@brynnz22 and I are [having trouble just accessing the swagger interface for submitting a changesheet.](https://github.com/microbiomedata/nmdc-schema/issues/1223#issuecomment-1786005047)

@dwinston or @eecavanna can you please look into this or delegate it to somebody else?

dwinston:
fixes #337

### DIFF
--- a/requirements/main.txt
+++ b/requirements/main.txt
@@ -6,10 +6,12 @@
 #
 alabaster==0.7.13
     # via sphinx
-alembic==1.12.0
+alembic==1.12.1
     # via dagster
 aniso8601==9.0.1
     # via graphene
+annotated-types==0.6.0
+    # via pydantic
 antlr4-python3-runtime==4.9.3
     # via
     #   linkml
@@ -29,9 +31,9 @@ argon2-cffi==23.1.0
     # via jupyter-server
 argon2-cffi-bindings==21.2.0
     # via argon2-cffi
-arrow==1.2.3
+arrow==1.3.0
     # via isoduration
-asttokens==2.4.0
+asttokens==2.4.1
     # via stack-data
 async-lru==2.0.4
     # via jupyterlab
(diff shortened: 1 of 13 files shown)


requirements/main.txt
```diff
@@ -549,7 +550,7 @@ pyasn1==0.5.0
     #   rsa
 pycparser==2.21
     # via cffi
-pydantic==1.10.13
+pydantic==2.4.2
```
dwinston, line 553: @turbomam this may require some attention wrt linkml-generated pydantic classes. raise as issue in nmdc-schema repo?

eecavanna: In case updating Pydantic from v1 to v2 here in `nmdc-runtime` leads to issues (i.e. making it difficult to use things generated by `nmdc-schema` in the short term), we may have another option: "patching" part of FastAPI so that it loads a specific version of the Swagger UI code (rather than updating the _whole_ of FastAPI, which implies also updating the Pydantic version).

Here's what such a "patch" looked like back in 2020: https://github.com/tiangolo/fastapi/issues/1762#issuecomment-662162612

eecavanna (approved):

PeopleMakeCulture (approved):
lgtm

requirements/main.txt
turbomam: @dwinston yes, unfortunately LInkML does not play nicely with pydantic 2 yet. I think there's active development in that area. @pkalita-lbl can you elaboarte?

pkalita-lbl: LinkML v1.6.0 added a flag to the Pydantic generator that allows you to switch between generating v1-compatible models and v2-compatible models. However that implementation had some bugs that are already fixed but awaiting release. The `main` branch of LinkML is fully tested under Pydantic 1 and 2. If it's urgently needed I'm sure we could get a new LinkML release out shortly.

But I'm not even sure where `nmdc-runtime` is using LinkML-generated Pydantic models. Unless I'm missing something it looks to me like it always uses LinkML-generated `@dataclass`-es

dwinston: there's just this one point of usage now: https://github.com/microbiomedata/nmdc-runtime/blob/4d5b5d7a3cdff4033ee5c3af5cfa5aedfd9a2619/components/nmdc_runtime/workflow_execution_activity/spec.py#L12 , which appears to be an unholy mixture of dataclass and pydantic model. perhaps we should proactively refactor that. was written by @elais (no blame! just "I only have a surface understanding of this area of the code").

pkalita-lbl: That looks like a hand-made Pydantic model where one of the fields is typed as a LinkML-generated `@dataclass`. Design choices aside, I don't think there's anything technically illegal about that, and I wouldn't be concerned about Pydantic v1/v2 compatibility there.

PR #347, opened 2023-11-02
fix: de-duplicate metadata submission after one minute

### LINKED ISSUE #340: Checksum Matches Existing Object - Json submit endpoint
When trying to re-submit a JSON that was previously deleted from production, I encounter the error:

`detail": "provided checksum matches existing object: {'checksums.type': 'sha256', 'checksums.checksum': '9eaece59f58b4d58008f2efcf556736c0f5a0cbe16311206d7f6b127e9cfc580'}`

dwinston:
re-submission of "same" changes is a valid use case

closes #340

### DIFF
--- a/nmdc_runtime/api/endpoints/util.py
+++ b/nmdc_runtime/api/endpoints/util.py
@@ -2,12 +2,14 @@ import logging
 import os
 import re
 import tempfile
+from datetime import datetime
 from functools import lru_cache
 from json import JSONDecodeError
 from pathlib import Path
 from time import time_ns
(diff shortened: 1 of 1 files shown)


eecavanna (approved):
Looks like this is to guarantee that the `base` value changes at least once per minute. I don't know the context here, but the code change looks straightforward to me.

eecavanna:
I recommend holding off on merging this until there is no risk of such a merge complicating the resolution of issue https://github.com/microbiomedata/nmdc-runtime/issues/348.

PeopleMakeCulture (approved):

Issue #350, opened 2023-11-03
API endpoint `/queries:run` returns HTTP 500 Internal Server Error


eecavanna:
Swagger UI link:
https://api.microbiomedata.org/docs#/queries/run_query_queries_run_post

The request payload is:

```json
{
"find": "biosample_set",
"filter": {"gold_biosample_identifiers": {"$elemMatch":{"$eq": "gold:Gb0356184"}}}
}
```

Here's a screenshot of the error response on Swagger UI:

![image](https://github.com/microbiomedata/nmdc-runtime/assets/134325062/d93b9185-5bc3-4db0-8ca8-382051c0d27f)

The final few lines of the stack trace on the server are:

```py
  File "/code/nmdc_runtime/api/endpoints/queries.py", line 79, in run_query
    cmd_response = _run_query(query, mdb)
  File "/code/nmdc_runtime/api/endpoints/queries.py", line 141, in _run_query
    mdb.query_runs.insert_one(query_run.model_dump(mode="json", exclude_unset=True))
  File "/usr/local/lib/python3.10/site-packages/pydantic/main.py", line 308, in model_dump
    return self.__pydantic_serializer__.to_python(
pydantic_core._pydantic_core.PydanticSerializationError: Unable to serialize unknown type: <class 'bson.objectid.ObjectId'>
```

Full stack trace is on Slack, here: 
https://nmdc-group.slack.com/archives/C05966WE4G6/p1698979668327079?thread_ts=1698960484.010949&cid=C05966WE4G6

Originally reported by @mbthornton-lbl on Slack, here:
https://nmdc-group.slack.com/archives/C05966WE4G6/p1698960484010949

CC: @PeopleMakeCulture

dwinston:
fixed by #363

PR #354, opened 2023-11-03
529 add ncbi biosample and project ids to neon soil

mbthornton-lbl:
This PR provides:

A python script to:
1. generate a changesheet to insert GOLD NcbiBioSampleAccessions and NcbiBioProjectAccessions to Biosample.insdc_biosample_identifiers and OmicsProcessing.insdc_experiment_identifiers.
2. validate the changes against the metadata/changesheets:validate endpoint.

Note: Validation was performed against the api-napa instance, dev and prod as of Nov 3 blocked by:
https://github.com/microbiomedata/nmdc-runtime/issues/350

Minor updates to simplify setup and some of the methods on the runtime API client.

Output Changefile:
https://drive.google.com/file/d/1bqBgUI8YfQT5ho0Lz02E2mzfCsaBKt-t/view?usp=drive_link


pkalita-lbl (approved):
Overall looks good. Two non-critical comments:

1. I guess I'm not fully up-to-speed on what the api-napa instance is, but I'm a little confused about the code that sort of uses that term interchangeably with "dev instance". To me "dev" is what's deployed on `api-dev.microbiomedata.org`. It might make more sense to me to remove the `get_runtime_client` method and the `--use_dev_api` flag, and let it all be driven by environment variables. Want to use the api-napa instance? Okay, then set your `API_HOST` envar accordingly. That's just my outsider view of how I'd expect something like this to work. If that doesn't make sense to you, forget I said anything.
2. It doesn't look like `python-dotenv` is listed in `main.in` but we get it as a transitive dependency of other stuff. If we're using it directly we should list it as a direct dependency.

pkalita-lbl (approved):

pkalita-lbl (approved):

PR #370, opened 2023-11-08
Update `nmdc-schema` package from `9.0.4` to `9.1.0`

eecavanna:
There is no data migration associated with (or necessary for / written for) this schema change.


eecavanna:
🤠 Yeehaw! Merging without review as part of tonight's "after-hours" series of migrations/updates to the dev environment.

PR #371, opened 2023-11-08
add checksum timestamp changes

brynnz22:
https://github.com/microbiomedata/nmdc-runtime/issues/340

Hi Donny, I am unfamiliar with the `hashlib` library, so I'm not sure about the code I added to the `sha256hash_from_file` function. I think this is what you meant.

Thanks!

### DIFF
--- a/nmdc_runtime/api/core/util.py
+++ b/nmdc_runtime/api/core/util.py
@@ -28,9 +28,12 @@ def hash_from_str(s: str, algo="sha256") -> str:
     return getattr(hashlib, algo)(s.encode("utf-8")).hexdigest()
 
 
-def sha256hash_from_file(file_path: str):
+def sha256hash_from_file(file_path: str, timestamp: str):
     # https://stackoverflow.com/a/55542529
     h = hashlib.sha256()
+
+    timestamp_bytes = timestamp.encode('utf-8')
+    h.update(timestamp_bytes)
 
     with open(file_path, "rb") as file:
         while True:
--- a/nmdc_runtime/api/endpoints/util.py
+++ b/nmdc_runtime/api/endpoints/util.py
@@ -442,6 +442,7 @@ def persist_content_and_get_drs_object(
                     ),
                     "access_methods": [{"access_id": drs_id}],
                 },
+                timestamp= datetime.now(tz=ZoneInfo('America/Los_Angeles')).isoformat(timespec='minutes')
             )
         )
     self_uri = f"drs://{HOSTNAME_EXTERNAL}/{drs_id}"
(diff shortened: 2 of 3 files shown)


nmdc_runtime/api/core/util.py
```diff
@@ -28,10 +28,13 @@ def hash_from_str(s: str, algo="sha256") -> str:
     return getattr(hashlib, algo)(s.encode("utf-8")).hexdigest()
 
 
-def sha256hash_from_file(file_path: str):
+def sha256hash_from_file(file_path: str, timestamp: str):
     # https://stackoverflow.com/a/55542529
     h = hashlib.sha256()
 
+    timestamp_bytes = timestamp.encode('utf-8')
+    h.update(timestamp_bytes)
```
eecavanna, line 36: This change makes sense to me.

[Docs](https://docs.python.org/3/library/stdtypes.html#str.encode) for `str.encode()` say that it returns the `bytes` equivalent of the string.

[Docs](https://docs.python.org/3/library/hashlib.html#hashlib.hash.update) for `hash.update()` say that it adds `bytes` to the hash (and that calling it multiple times is like concatenating the individual arguments and calling it once with the concatenation result).

nmdc_runtime/api/endpoints/util.py
```diff
@@ -442,6 +442,7 @@ def persist_content_and_get_drs_object(
                     ),
                     "access_methods": [{"access_id": drs_id}],
                 },
+                timestamp= datetime.now(tz=ZoneInfo('America/Los_Angeles')).isoformat(timespec='minutes')
```
eecavanna, line 445: This change makes sense to me. 

A `string` is being generated that indicates the current time (in Los Angeles, CA) down to the minute.

[Docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat) for `datetime.isoformat()` say it returns a `string` representation of the time in ISO 8601 format.

image

eecavanna: I recommend removing the space after the equals sign, since this is "argument setting" as opposed to "variable assignment" (I think that's a [Python convention](https://peps.python.org/pep-0008/#other-recommendations)). 

```diff
- timestamp= datetime.now(...
+ timestamp=datetime.now(...
```

eecavanna (approved):
The code changes, in isolation, make sense to me. I will defer to @dwinston for final approval and merging, as I'm not too familiar with the context (i.e. this part of the code base) and I think these changes stemmed from a comment he made in https://github.com/microbiomedata/nmdc-runtime/issues/340#issuecomment-1802228546.

dwinston (approved):

PR #377, opened 2023-11-13
fix: coerce pydantic AnyUri to str when needed

### LINKED ISSUE #376: Pydantic `AttributeError` (on Dagit) after using `/metadata/json:submit` endpoint
When trying to submit a json using the json:submit endpoint, the job failed. The error looks like: 

Screenshot 2023-11-13 at 10 05 32 AM

@aclum

dwinston:
fixes #376

### DIFF
--- a/nmdc_runtime/site/resources.py
+++ b/nmdc_runtime/site/resources.py
@@ -17,7 +17,7 @@ from dagster import (
 from fastjsonschema import JsonSchemaValueException
 from frozendict import frozendict
 from linkml_runtime.dumpers import json_dumper
-from pydantic import BaseModel
+from pydantic import BaseModel, AnyUrl
 from pymongo import MongoClient, ReplaceOne, InsertOne
 from terminusdb_client import WOQLClient
 from toolz import get_in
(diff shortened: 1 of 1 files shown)


nmdc_runtime/site/resources.py
```diff
@@ -194,15 +194,17 @@ def get_object_bytes(self, object_id) -> requests.Response:
             access = AccessURL(
                 **self.get_object_access(object_id, method.access_id).json()
             )
-            if access.url.startswith(
+            if str(access.url).startswith(
```
eecavanna, line 197: Side note: Maybe there's already a way to get the `url` property of `access` as a string, rather than casing it here. I noticed the `AccessURL` class has this method defined (I don't know what the decorator does yet).

https://github.com/microbiomedata/nmdc-runtime/blob/7bc8f4066898a73863700b12b3d1e66abf31e5bc/nmdc_runtime/api/models/object.py#L31-L37

eecavanna:
LGTM! I posted a comment that I do not consider a blocker for merge.

eecavanna (approved):

nmdc_runtime/site/resources.py
```diff
@@ -194,15 +194,17 @@ def get_object_bytes(self, object_id) -> requests.Response:
...
-                    os.getenv("API_HOST_EXTERNAL"), os.getenv("API_HOST")
+                access.url = AnyUrl(
+                    str(access.url).replace(
+                        os.getenv("API_HOST_EXTERNAL"), os.getenv("API_HOST")
+                    )
                 )
         else:
             access = AccessURL(url=method.access_url.url)
-        return requests.get(access.url)
+        return requests.get(str(access.url))
```
eecavanna, line 207: Side note: Instead of casting it 3x, we could store the casted result as `access_url_str`.

nmdc_runtime/site/resources.py
dwinston: the `field_serializer` decorator ensures that when `.model_dump()` is called on a model, `url` is stringified, but this doesn't hold for direct attribute access, i.e. `<model>.url`.

nmdc_runtime/site/resources.py
dwinston: yeah, I considered that. we get some url (format) validation by keeping to the model until the return.

eecavanna:
Thanks @dwinston !

FYI @brynnz22 , here's the web page I look at to see whether code that was "just now" merged into `main` has been deployed to the production environment yet:

https://github.com/microbiomedata/nmdc-runtime/actions

image

Once both of those workflow runs has a green checkmark next to it, the code has been deployed to both the development and production environments.

PR #382, opened 2023-11-16
Separate dev and production deployments in GitHub workflow

### LINKED ISSUE #357: Reconfigure GitHub Actions to have separate dev and production deployments
Currently we have:

* A `build-and-push-docker-images` [workflow](https://github.com/microbiomedata/nmdc-runtime/blob/dc668eaa9efb210336947673a9df379ccaf2b9f1/.github/workflows/build-and-push-docker-images.yml) which runs either manually or automatically on pushes to `main`. As the name suggests, it builds Docker images and pushes them to Docker Hub.
* A `release-to-spin` [workflow](https://github.com/microbiomedata/nmdc-runtime/blob/dc668eaa9efb210336947673a9df379ccaf2b9f1/.github/workflows/release-to-spin.yml) which is triggered by the successful completion of the `build-and-push-docker-images` workflow. This workflow issues Rancher CLI commands (via [this action](https://github.com/colbyhill21/Rancher-Action)) to restart the `runtime-api`, `dagster-dagit`, and `dagster-daemon` deployments on dev and production. These deployments all pull the `main`-tagged images.

The effect of this is that PR merges (or direct pushes to `main`) are immediately deployed to both dev and production. This is a double-edged sword. We can quickly get backwards-compatible fixes out quickly, but there's also no guard against backwards-incompatible changes going out before downstream clients are ready.

The proposal here is to move to a model where:

* Deployments to dev happen immediately on pushes to `main`, as happens today
* Deployments to production are pinned to tagged release versions

In a perfect world the only manual action needed to do a production deployment would be creating a GitHub release and everything else would happen automatically. That might not be possible at the moment because of manual operations needed for data migration when the `nndc-schema` dependency changes. We should automate as much as possible and [clearly document](https://github.com/microbiomedata/infra-admin/issues/19) the rest.

This task is to do any updates to workflows in this repo and configuration in Racher to support deploying tagged versions to production.

pkalita-lbl:
These changes consolidate the old `build-and-push-docker-images.yml` and `release-to-spin.yml` workflows into one new one (`build-and-release-to-spin.yml`). The new workflow is responsible for building the two docker images, pushing them to Docker Hub with the correct tags, and redeploying Spin workloads. The Docker image tags are generated by the [`docker/metadata-action`](https://github.com/docker/metadata-action). 

When the new workflow runs on a push to the `main` branch, it will push Docker images with the `main` tag and then redeploy workloads in the `nmdc-dev` namespace. When the workflow runs on a push to a tag like `vX.Y.Z` it will push docker images with the `X.Y.Z` tag _and_ the `prod` tag and then redeploy workloads in the `nmdc` namespace. Rancher will need to be manually updated so that the `nmdc` namespace workloads pull the `prod`-tagged image.

To avoid confusion I completely removed the now unused `docker-build.sh` script which tagged images with commit hashes. I lightly reformatted part of the Release Process doc page which was still referring to actions being triggered on changes to `RELEASES.md` (that hasn't been accurate for a while now). I intend to do more in-depth documentation with https://github.com/microbiomedata/infra-admin/issues/19.

Fixes #357


.github/workflows/build-and-release-to-spin.yml
```diff
@@ -0,0 +1,81 @@
+name: Build Docker images and release to Spin
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'v*'
```
dwinston, line 8: ooh, I didn't know about this wildcard syntax. cool.

docs/howto-guides/release-process.md
```diff
@@ -15,16 +15,14 @@ released? Here's how.
...
      action](https://github.com/microbiomedata/nmdc-runtime/blob/main/.github/workflows/python-app.yml),
      which is triggered automatically when a change to any Python file in the repository is pushed to the
-     `main` branch, or to a Pull Request. You can monitor the status of Github Actions
+     `main` branch, or to a Pull Request. You can monitor the status of GitHub Actions
      [here](https://github.com/microbiomedata/nmdc-runtime/actions).
 
-2. Add a summary for the release to `RELEASES.md`. You can make an edit and push to the `main`
-   branch [via GitHub](https://github.com/microbiomedata/nmdc-runtime/blob/main/RELEASES.md). This will
-   trigger two GitHub actions in sequence to
+2. Create a new [GitHub Release](https://github.com/microbiomedata/nmdc-runtime/releases). This will trigger GitHub Actions which:
```
dwinston, line 21: add some qualification like "tag it by typing 'v' and then using the autocomplete to fill in a release tag of the form 'vX.Y.Z', where if the release is a patch (fixing/refactoring something without changing behavior), increment Z by 1 relative to the last-release tag, if the release is minor (adds some new behavior without changing existing behavior), increment Y by 1 and set Z to 0, or if the release is major (old behavior has changed, so may break existing consumer workflows), increment X by 1 and set both Y and Z to 0."?

dwinston (changes requested):
seems good, apart from adding some more guidance around release naming.

docs/howto-guides/release-process.md
pkalita-lbl: Done.

dwinston (approved):
great

PR #383, opened 2023-11-16
Updates to submission portal-to-Mongo pipeline to support GLBRC data

pkalita-lbl:
Re: https://github.com/microbiomedata/submission-schema/issues/162

These changes are all around support for bringing data from a GLBRC submission into Mongo. Summary of changes:

* In the latest version of the schema `dataset_dois` has been replaced by `associated_dois` on the `Study` class. The range of `associated_dois` is a `Doi` class which has two required slots and one conditionally required slot. These changes update the `SubmissionPortalTranslator` to adhere to that. However since the submission portal does not yet collect some of the newly required information, it is being piped in via Dagster `op` inputs instead. So, _for now_ the person running this pipeline will be responsible for filling in these values. 
* Similarly, `study_category` is now required on the `Study` class, but has no corresponding form field in the submission portal. These changes allow passing it in while running the pipeline.
* The study/PI image and study funding sources aren't new or updated in the schema. This is just something that has been overlooked by the submission portal for a while. Again, allowing those to be injected into the pipeline run.
* The GLBRC has provided data that's split between _two_ environment packages. That means they can't input it all in the submission portal. But it all ends up on `Biosample` in Mongo anyway. So these changes allow for injecting extra `Biosample` data via a remote CSV file along with a file mapping their column headers to NMDC slot names along with units.

### DIFF
--- a/nmdc_runtime/site/ops.py
+++ b/nmdc_runtime/site/ops.py
@@ -643,24 +643,27 @@ def nmdc_schema_database_from_gold_study(
 
 
 @op(
-    config_schema={
-        "submission_id": str,
-        "omics_processing_mapping_file_url": str,
-        "data_object_mapping_file_url": str,
-    },
     out={
         "submission_id": Out(),
-        "omics_processing_mapping_file_url": Out(),
-        "data_object_mapping_file_url": Out(),
+        "omics_processing_mapping_file_url": Out(Optional[str]),
+        "data_object_mapping_file_url": Out(Optional[str]),
+        "biosample_extras_file_url": Out(Optional[str]),
+        "biosample_extras_slot_mapping_file_url": Out(Optional[str]),
     },
 )
 def get_submission_portal_pipeline_inputs(
-    context: OpExecutionContext,
-) -> Tuple[str, str, str]:
+    submission_id: str,
+    omics_processing_mapping_file_url: Optional[str],
+    data_object_mapping_file_url: Optional[str],
+    biosample_extras_file_url: Optional[str],
+    biosample_extras_slot_mapping_file_url: Optional[str],
+) -> Tuple[str, str | None, str | None, str | None, str | None]:
     return (
-        context.op_config["submission_id"],
-        context.op_config["omics_processing_mapping_file_url"],
-        context.op_config["data_object_mapping_file_url"],
+        submission_id,
+        omics_processing_mapping_file_url,
+        data_object_mapping_file_url,
+        biosample_extras_file_url,
+        biosample_extras_slot_mapping_file_url,
     )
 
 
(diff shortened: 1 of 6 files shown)


nmdc_runtime/site/translation/submission_portal_translator.py
```diff
@@ -21,6 +21,34 @@ def _get_schema_view():
...
+def group_dicts_by_key(key: str, seq: Optional[list[dict]]) -> Optional[dict]:
+    """Transform a sequence of dicts into a single dict based on values of `key` in each dict.
+
+    Unlike toolz.groupby: 1) this method only applies to sequences of dicts, 2) it only keeps one
+    value for each key, 3) it removes the key from the original dict.
+
+    Example:
+        >>> seq = [{'a': 'one', 'b': 1}, {'a': 'two', 'b': 2}, {'a': 'two', 'b': 3}]
+        >>> group_dicts_by_key('a', seq)
+        {'one': {'b': 1}, 'two': {'b': 3}}
```
eecavanna, line 33: Super helpful doctest!

nmdc_runtime/site/ops.py
```diff
@@ -680,6 +683,13 @@ def translate_portal_submission_to_nmdc_schema_database(
     metadata_submission: Dict[str, Any],
     omics_processing_mapping: List,
     data_object_mapping: List,
+    study_category: Optional[str],
+    study_doi_category: Optional[str],
+    study_doi_provider: Optional[str],
+    study_funding_sources: Optional[List[str]],
+    study_pi_image_url: Optional[str],
+    biosample_extras: Optional[list[dict]],
+    biosample_extras_slot_mapping: Optional[list[dict]],
```
eecavanna, line 692: Note to self (@eecavanna):
- Consider that list typehints are `list` in some places and `List` (from `typing`) in other places
- `list` became usable as of some recent version of Python (making `typing.List` unnecessary)
- Look at what Python version this project's docs indicates it requires

nmdc_runtime/site/translation/submission_portal_translator.py
```diff
@@ -61,9 +121,10 @@ def _get_pi(
             name=study_form.get("piName"),
             email=study_form.get("piEmail"),
             orcid=study_form.get("piOrcid"),
+            profile_image_url=self.study_pi_image_url,
         )
 
-    def _get_doi(self, metadata_submission: JSON_OBJECT) -> Union[List[str], None]:
+    def _get_doi(self, metadata_submission: JSON_OBJECT) -> Union[List[nmdc.Doi], None]:
```
eecavanna, line 127: Another note to myself (@eecavanna):
- `nmdc.Doi` is a `class` defined in the `nmdc_schema` package (installed via PyPI)

eecavanna (approved):

PR #384, opened 2023-11-16
Misc. updates to `gold_translator.py`

sujaypatil96:
As a result of the discussion on https://github.com/microbiomedata/issues/issues/527 a couple of trivial edits needed to be made to the GOLD-to-Mongo ingest pipeline:
* Use lowercase `gold` while making CURIEs for GOLD entities
* Modification to the logic in `_get_quantity_value()` to accommodate special handling of _depth_ data

> NMDC depth.has_minimum_numeric_value — GOLD depthInMeters
> NMDC depth.has_maximum_numeric_value — GOLD depthInMeters2

> depthInMeters2 may or may not be present in GOLD, so in the cases where it’s not provided:
> NMDC depth.has_numeric_value — GOLD depthInMeters

### DIFF
--- a/nmdc_runtime/site/translation/gold_translator.py
+++ b/nmdc_runtime/site/translation/gold_translator.py
@@ -1,6 +1,6 @@
 import collections
 import re
-from typing import List, Union
+from typing import List, Tuple, Union
 from nmdc_schema import nmdc
 
 from nmdc_runtime.site.translation.translator import JSON_OBJECT, Translator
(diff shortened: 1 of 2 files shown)


nmdc_runtime/site/translation/gold_translator.py
```diff
@@ -210,11 +210,24 @@ def _get_quantity_value(
...
+        has_minimum_numeric_value = None
+        has_maximum_numeric_value = None
+
         # TODO: in the future we will need better handling
         # to parse out the numerical portion of the quantity value
         # ex. temp might be 3 C, and we will need to parse out 3.0 from it
         if unit == "meters":
             numeric_value = nmdc.Double(field_value)
+            
+            has_maximum_numeric_value = gold_entity.get("depthInMeters2")
```
pkalita-lbl, line 222: `_get_quantity_value` is supposed to be a generic method, so I don't like hardcoding `depthInMeters2` here. It also doesn't really make a lot of sense to me why we only possibly set `has_minimum_numeric_value` and `has_maximum_numeric_value` based on the `unit`?? Why not add parameters for `minimum_value_field` and `maximum_value_field`?

pkalita-lbl (changes requested):

nmdc_runtime/site/translation/gold_translator.py
sujaypatil96: Yup, I was contemplating the specificity of this solution too 😫 I'll make the logic more generic keeping your suggestions in mind, and add another commit to this PR!

If not accepting `minimum_value_field` and `maximum_value_field` as arguments of `_get_quantity_value()` we can also allow the `gold_field` argument to accept `Union[str, List[str]]` and if it gets a list of two items, then use the first value/field in that list to get the minimum value, and the second value/field in that list to get the maximum value.

pkalita-lbl: That makes sense too! Although if you go the second route use `Tuple[str, str]` not `List[str]` to indicate that it's not an arbitrary length list, but rather a pair of strings.

sujaypatil96: That's a great suggestion! I used `Tuple[str, str]`.

pkalita-lbl (approved):

PR #389, opened 2023-11-20
Replace illegal variable defined using other variable

pkalita-lbl:
Follow up to #382. 

Apparently variables cannot be define in terms of other variables in GitHub workflows. This changes the main release workflow so that all the variables are defined up top and shared with downstream jobs using [environment files](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files).

### DIFF
--- a/.github/workflows/build-and-release-to-spin.yml
+++ b/.github/workflows/build-and-release-to-spin.yml
@@ -17,6 +17,14 @@ env:
   IS_PROD_RELEASE: ${{ startsWith(github.ref, 'refs/tags/v') }}
 
 jobs:
+  setup-variables:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set variables
+        run: |
+          echo "IS_PROD_RELEASE=${{ startsWith(github.ref, 'refs/tags/v') }}" >> "$GITHUB_ENV"
+          echo "RANCHER_NAMESPACE=${{ startsWith(github.ref, 'refs/tags/v') && 'nmdc' || 'nmdc-dev' }}" >> "$GITHUB_ENV"
+
   build:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-release-to-spin.yml
```diff
@@ -70,11 +78,9 @@ jobs:
         deployment: [ runtime-api, dagster-dagit, dagster-daemon ]
 
     steps:
-      - name: Set env var
-        run: echo "NAMESPACE=${{ env.IS_PROD_RELEASE && 'nmdc' || 'nmdc-dev' }}" >> "$GITHUB_ENV"
-      - name: Redeploy ${{ env.NAMESPACE }}:${{ matrix.deployment }}
+      - name: Redeploy ${{ env.RANCHER_NAMESPACE }}:${{ matrix.deployment }}
```
dwinston, line 81: yeah, good call taking `NAMESPACE` to `RANCHER_NAMESPACE`.

dwinston (approved):
thanks

.github/workflows/build-and-release-to-spin.yml
```diff
@@ -17,6 +17,14 @@ env:
   IS_PROD_RELEASE: ${{ startsWith(github.ref, 'refs/tags/v') }}
 
 jobs:
+  setup-variables:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set variables
+        run: |
+          echo "IS_PROD_RELEASE=${{ startsWith(github.ref, 'refs/tags/v') }}" >> "$GITHUB_ENV"
+          echo "RANCHER_NAMESPACE=${{ startsWith(github.ref, 'refs/tags/v') && 'nmdc' || 'nmdc-dev' }}" >> "$GITHUB_ENV"
```
eecavanna, line 26: Post-merge note to myself:

`startsWith` here is a GitHub Actions Expression: https://docs.github.com/en/actions/learn-github-actions/expressions#startswith

`&&` is an operator: https://docs.github.com/en/actions/learn-github-actions/expressions#operators

This "ternary operator"-like behavior is documented in the following example in the GitHub Actions Expressions documentation: https://docs.github.com/en/actions/learn-github-actions/expressions#example

PR #390, opened 2023-11-20
Do full depth checkout so setuptools-scm can detect version

pkalita-lbl:
Another follow up to #382


eecavanna (approved):
Reference to `fetch-depth` in the `checkout` action's docs: https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches

PR #391, opened 2023-11-21
Revert to using env context for variables

pkalita-lbl:
Follow up to #390.

I thought I understood the GitHub workflows environment file thing, but oops not quite. I think those values are only sent to the runners so they can't be used for things like `enable=${{ env.IS_PROD_RELEASE }}` outside of `run` in a `step`. So, back to using a workflow-level `env` block.


eecavanna (approved):
LGTM!

To me, this approach seems consistent with an example I saw in the GitHub docs earlier: 
https://docs.github.com/en/actions/learn-github-actions/expressions#example

PR #404, opened 2023-11-27
Mix in orcid jwt flow with client_credentials flow

### LINKED ISSUE #333: use ORCiD for authn
This ticket consists of two tasks:

1. Create a new API endpoint that calls ORCID's auth url and returns a Token

> Options:
> - the most secure is a two-step process, using an authorization code to request a token
> - alternatively, we could use the implicit flow

2. Add a new authentication flow to the FastAPI OAuth2 model that uses the ORCID access token

- add `openid_oauth2_scheme` here using the authorizationCode flow: https://github.com/microbiomedata/nmdc-runtime/blob/ebe41c9874ccf0326b925462afd6f8f672b6671e/nmdc_runtime/api/core/auth.py#L116
(alt: we could also use the implicit openid flow, as described in the orcid dev tools)

DOCS: [fastapi.security.OAuthFlows.authorizationCode](https://fastapi.tiangolo.com/reference/openapi/models/#fastapi.openapi.models.OAuthFlows.authorizationCode)

dwinston:
User can load `/orcid_authorize` to obtain orcid JWT. Then authorize using `client_credentials` flow using orcid jwt as `client_id` and empty `client_secret`. Will create user with orcid as username if it doesn't exist. Returns a user token just like the username/password flow.

closes #333


nmdc_runtime/api/core/auth.py
```diff
@@ -99,17 +111,21 @@ async def __call__(self, request: Request) -> Optional[str]:
         scheme, param = get_authorization_scheme_param(authorization)
         if not authorization or scheme.lower() != "bearer":
             if self.auto_error:
+                print("uh oh")
```
eecavanna, line 114: Marking this as something someone may want to revisit (i.e. elaborate on, remove) as part of the upcoming "simplify auth" work.

nmdc_runtime/api/endpoints/users.py
```diff
@@ -22,6 +32,40 @@
...
+    """NOTE: You want to load /orcid_authorize directly in your web browser to initiate the login redirect flow."""
+    return RedirectResponse(
+        f"https://orcid.org/oauth/authorize?client_id={ORCID_CLIENT_ID}"
+        "&response_type=token&scope=openid&"
+        f"redirect_uri={BASE_URL_EXTERNAL}/orcid_token"
+    )
+
+
+@router.get("/orcid_token")
+async def redirect_uri_for_orcid_token(req: Request):
```
eecavanna, line 46: Add docstring:

```py
r"""
Returns the HTML of a web page that...
"""
```

nmdc_runtime/api/core/auth.py
```diff
@@ -16,6 +16,18 @@
 
 SECRET_KEY = os.getenv("JWT_SECRET_KEY")
 ALGORITHM = "HS256"
+ORCID_CLIENT_ID = os.getenv("ORCID_CLIENT_ID")
+
+# https://orcid.org/.well-known/openid-configuration
+# XXX do we want to live-load this?
```
eecavanna, line 22: If the JSON content at https://orcid.org/oauth/jwks is something that changes over time (I don't know whether it is), I'd prefer to live-load it.

eecavanna (dismissed):
I left a few comments about documentation and print statements.

dwinston:
@eecavanna great feedback, thanks.

nmdc_runtime/api/core/auth.py
dwinston: hmm. yeah, will include in refactoring.

PR #409, opened 2023-11-30
Allow Biosample.env_package to be populated from submission-level data

pkalita-lbl:
This is part 2 of https://github.com/microbiomedata/issues/issues/543.

The changes here update the `SubmissionPortalTranslator` class so so that when it is creating `Biosample` instances, the `env_package` slot is populated by default from the submission-level `packageName` field. Tests updated accordingly.


pkalita-lbl:
Ping @dwinston @PeopleMakeCulture @eecavanna. Still looking for review. Can anyone take a look?

eecavanna (approved):
I think I overlooked this PR originally. I reviewed it just now and, although I don't know much about the context, all the changes in this branch do look intentional to me. LGTM.

PR #411, opened 2023-12-04
allow mongodb update command via POST /query:run

### LINKED ISSUE #356: revisit endpoints used to update mongo data
I'd like to revisit alternatives to changesheets, for example, perhaps an endpoint which would take pymongo find and update commands directly.
To be discussed an an upcoming infrastructure sync

cc @shreddd @eecavanna @dwinston @mbthornton-lbl @brynnz22 @mslarae13 @ssarrafan

dwinston:
closes #356

#### A successful `update` query run 
- sends a 200
image
- updates the document
image

#### A poorly formatted `update` query run
- Throws an error if no records found
image

- Throws an error if query is poorly formatted
image

### DIFF
--- a/nmdc_runtime/api/endpoints/queries.py
+++ b/nmdc_runtime/api/endpoints/queries.py
@@ -29,11 +29,13 @@ def unmongo(d: dict) -> dict:
     return json.loads(bson.json_util.dumps(d))
 
 
-def check_can_delete(user: User):
+def check_can_update_and_delete(user: User):
+    # update and delete queries require same level of permissions
     if not permitted(user.username, "/queries:run(query_cmd:DeleteCommand)"):
         raise HTTPException(
             status_code=status.HTTP_403_FORBIDDEN,
-            detail=(f"Only specific users " "are allowed to issue delete commands.",),
+            detail=(f"Only specific users " 
+                    "are allowed to issue update and delete commands.",),
         )
 
 
@@ -64,7 +66,7 @@ def run_query(
     ```
     """
     if isinstance(query_cmd, (DeleteCommand, UpdateCommand)):
-        check_can_delete(user)
+        check_can_update_and_delete(user)
 
     qid = generate_one_id(mdb, "qy")
     saved_at = now()
@@ -102,7 +104,7 @@ def rerun_query(
     doc = raise404_if_none(mdb.queries.find_one({"id": query_id}))
     query = Query(**doc)
     if isinstance(query, DeleteCommand):
-        check_can_delete(user)
+        check_can_update_and_delete(user)
 
     cmd_response = _run_query(query, mdb)
     return unmongo(cmd_response.model_dump(exclude_unset=True))
--- a/nmdc_runtime/api/endpoints/queries.py
+++ b/nmdc_runtime/api/endpoints/queries.py
@@ -65,7 +65,7 @@ def run_query(
 
     {
         "update": "biosample_set",
-        "updates": [{"q": {"id": "nmdc:bsm-11-002vgm56"}, "u": {"$set": {"name": "BEAL_048-O-6.5-19.5-20160725"}}}]
+        "updates": [{"q": {"id": "YOUR_BIOSAMPLE_ID"}, "u": {"$set": {"name": "A_NEW_NAME"}}}]
     }
     ```
     """
--- a/nmdc_runtime/api/endpoints/queries.py
+++ b/nmdc_runtime/api/endpoints/queries.py
@@ -113,7 +113,7 @@ def _run_query(query, mdb) -> CommandResponse:
     ran_at = now()
     if q_type is DeleteCommand:
         collection_name = query.cmd.delete
-        if collection_name not in nmdc_schema_collection_names():
+        if collection_name not in nmdc_schema_collection_names(mdb):
             raise HTTPException(
                 status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
                 detail="Can only delete documents in nmdc-schema collections.",
@@ -135,7 +135,7 @@ def _run_query(query, mdb) -> CommandResponse:
                 )
     elif q_type is UpdateCommand:
         collection_name = query.cmd.update
-        if collection_name not in nmdc_schema_collection_names():
+        if collection_name not in nmdc_schema_collection_names(mdb):
             raise HTTPException(
                 status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
                 detail="Can only update documents in nmdc-schema collections.",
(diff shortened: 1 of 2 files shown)


nmdc_runtime/api/endpoints/queries.py
```diff
@@ -135,5 +163,14 @@ def _run_query(query, mdb) -> CommandResponse:
         if cmd_response.ok
         else QueryRun(qid=query.id, ran_at=ran_at, error=cmd_response)
     )
+    if q_type in (DeleteCommand, UpdateCommand) and cmd_response.n == 0:
+        raise HTTPException(
+            status_code=status.HTTP_418_IM_A_TEAPOT,
+            detail=(
+                f"{'update' if q_type is UpdateCommand else 'delete'} command modified zero documents."
+                " I'm guessing that's not what you expected. Check the syntax of your request."
+                " But what do I know? I'm just a teapot.",
```
eecavanna, line 172: 😆

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -127,6 +133,28 @@ def _run_query(query, mdb) -> CommandResponse:
...
                 )
+    elif q_type is UpdateCommand:
+        collection_name = query.cmd.update
+        if collection_name not in nmdc_schema_collection_names(mdb):
+            raise HTTPException(
+                status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                detail="Can only update documents in nmdc-schema collections.",
+            )
+        update_specs = [
+            {"filter": us.q, "limit": 0 if us.multi else 1} for us in query.cmd.updates
```
eecavanna, line 144: I am not seeing where `us` is defined. I may be having a brain 💨 over here...

eecavanna, line 144: I don't understand what the `.multi` property does yet. This is after having looked at the `class UpdateStatement(BaseModel):` class definition. I'll keep reading—maybe it will become more clear to me.

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -127,6 +133,28 @@ def _run_query(query, mdb) -> CommandResponse:
...
                     detail="Failed to back up to-be-deleted documents. operation aborted.",
                 )
+    elif q_type is UpdateCommand:
+        collection_name = query.cmd.update
+        if collection_name not in nmdc_schema_collection_names(mdb):
+            raise HTTPException(
+                status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                detail="Can only update documents in nmdc-schema collections.",
+            )
+        update_specs = [
```
eecavanna, line 143: I don't understand why this is named `update_specs` while the (pre-existing) one for the `DeleteCommand` is named `find_specs`. I think they are equally "find" spec (i.e. finding the thing we want to either update or delete).

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -62,7 +63,7 @@ def run_query(
     }
     ```
     """
-    if isinstance(query_cmd, DeleteCommand):
+    if isinstance(query_cmd, (DeleteCommand, UpdateCommand)):
         check_can_delete(user)
```
eecavanna, line 67: I think there is an inconsistence between the name of the function being called here (i.e. `..._delete`) and the conditions for calling it.

nmdc_runtime/api/endpoints/queries.py
PeopleMakeCulture: if `update_specs.multi` is `True` the update will apply to all query results for the update query. (eg `limit: 0`)  

image

docs: https://www.mongodb.com/docs/v6.0/reference/command/update/#std-label-update-command-output
see also: https://www.mongodb.com/docs/v6.0/reference/command/update/#std-label-multi-update-failures

nmdc_runtime/api/endpoints/queries.py
PeopleMakeCulture: Good catch! We did not create a `check_can_update`, since it would apply the same rules as `check_can_delete`. I'll document this in the code.

nmdc_runtime/api/models/query.py
```diff
@@ -89,6 +90,32 @@ class DeleteCommandResponse(CommandResponse):
     writeErrors: Optional[List[Document]] = None
 
 
+class UpdateStatement(BaseModel):
+    q: Document
+    u: Document
+    upsert: bool = False
+    multi: bool = False
```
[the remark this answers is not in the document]
PeopleMakeCulture: @eecavanna do you mean the `multi` property?

nmdc_runtime/api/endpoints/queries.py
PeopleMakeCulture: I updated `find_specs` to `delete_specs` for consistency and updated a few var names for clarity

nmdc_runtime/api/models/query.py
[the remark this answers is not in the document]
eecavanna: Yes, the `multi` property.

Thanks for linking to the Mongo docs in the other comment. I think this comment (for example) would suffice:

```diff
    q: Document
    u: Document
    upsert: bool = False
+   # If true, all documents that meet the query criteria will be updated.
+   # Otherwise, only a single document that meets the query criteria will be updated.
    multi: bool = False
```

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -29,11 +29,13 @@ def unmongo(d: dict) -> dict:
...
 
-def check_can_delete(user: User):
+def check_can_update_and_delete(user: User):
+    # update and delete queries require same level of permissions
     if not permitted(user.username, "/queries:run(query_cmd:DeleteCommand)"):
         raise HTTPException(
             status_code=status.HTTP_403_FORBIDDEN,
-            detail=(f"Only specific users " "are allowed to issue delete commands.",),
+            detail=(f"Only specific users " 
+                    "are allowed to issue update and delete commands.",),
```
eecavanna, line 38: I think the space between the two strings is a remnant of a recent change (removing the username list). The strings can be combined into a single one.

```diff
-            detail=(f"Only specific users " 
-                   "are allowed to issue update and delete commands.",),
+             detail=(f"Only specific users are allowed to issue update and delete commands."),
```

Caveat: `black` may still split the single string onto multiple lines due to line length.

eecavanna (approved):
I added two comments about strings. I don't consider either one to be a deal breaker (i.e. merge blocker).

nmdc_runtime/api/models/query.py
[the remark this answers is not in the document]
PeopleMakeCulture: Done

PeopleMakeCulture (approved):
lgtm!

PR #418, opened 2023-12-13
fix: unique index on username for users collection

### LINKED ISSUE #325: Database allows multiple `users` documents to have the same `username` value
The Mongo database allows for there to be multiple `users` documents that have the same `username` value as one another.

Meanwhile, the `_runtime.api.allow` collection uses `username` as a key. So, it is not always clear to me which `users` document a given `_runtime.api.allow` document is "referring" to.

For example, the dev database has two users named `mbtho...` [real username truncated here]

I propose introducing a "unique index" of the `username` field.

dwinston:
closes #325

note: I just built an index on each of dev and prod, so deploying this involves no risk of failure to startup due to failure of index creation.

### DIFF
--- a/nmdc_runtime/api/main.py
+++ b/nmdc_runtime/api/main.py
@@ -249,7 +249,7 @@ def ensure_initial_resources_on_boot():
             ).model_dump(exclude_unset=True),
             upsert=True,
         )
-        mdb.users.create_index("username")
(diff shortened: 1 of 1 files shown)


eecavanna (approved):
LGTM! Thanks for fixing this.

PR #430, opened 2024-01-05
Pipeline to ingest data from NEON surface water data product

sujaypatil96:
Fixes https://github.com/microbiomedata/issues/issues/334


sujaypatil96:
@pkalita-lbl this code works and is producing JSON as expected. Should we merge this in? I'm not sure if it's worth the effort to write up tests for it considering the JSON produced by the pipeline is already in dev and prod Mongo, and we won't be re-running the pipeline again.

pkalita-lbl (approved):
The changes seems fine.

I was looking at the linked issue for more context, and I was a little surprised to see that it was already closed with a note saying that the data had already been ingested. The fact that reviewing and merging code is the _last_ thing to happen is a bit of an odd smell to me that makes me think something's gone awry in our process. We should dissect what's going on here some time (not necessarily in this PR).

PR #451, opened 2024-01-23
Update `nmdc-schema` from `v9.3.2` to `v10.0.0`

eecavanna:
### Summary of changes

- Implemented Jupyter/Python Notebook that migrates data from schema `v9.3.2` to `v10.0.0`
- Updated `nmdc-schema` dependency in `main.in`
  - Then, synchronized dependencies by running `$ make update-deps` within the `fastapi` container


eecavanna:
I will merge this in without review.

PR #453, opened 2024-01-24
DO NOT MERGE: Pin `nmdc-schema` to `7.8.0` for Napa Compliance squad's environment

eecavanna:
# ⚠️ DO NOT MERGE ⚠️ 

### Summary of changes

- [x] Pins `nmdc-schema` to `7.8.0`
- [x] Syncs dependencies accordingly (via `$ make update-deps`)

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -4,25 +4,12 @@
 #
 #    pip-compile --allow-unsafe --output-file=requirements/dev.txt --strip-extras requirements/dev.in
 #
-aiohttp==3.9.1
+attrs==23.2.0
     # via
     #   -c requirements/main.txt
-    #   black
-aiosignal==1.3.1
-    # via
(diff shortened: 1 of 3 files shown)


turbomam:
Could somebody explain why we're pinning to v7.8.0 even though it will fail gold identifiers with the wrong capitalizastion? Isn't v8.0.0 better?

eecavanna:
The reason I chose `nmdc-schema` version 7.8.0, specifically, is that that's the version of `nmdc-schema` the Napa Compliance squad has been using for the past few months and I didn't want to change it.

eecavanna:
I'm OK with this PR being closed. The reason I created it in the first place was that I didn't see a way to comment on the branch on GitHub, without it being in a PR; and I wanted to comment on it (specifically, about whether `$ make update-deps` had been run as of the initial commit).

Issue #437, opened 2024-01-11
co-locate as much API usage documentation with OpenAPI spec and thus on Swagger display


dwinston:

dwinston:
e.g. migrate content from https://github.com/microbiomedata/NMDC_documentation/blob/main/docs/howto_guides/api_gui.md to this repo

PeopleMakeCulture:
PR: https://github.com/microbiomedata/nmdc-runtime/pull/455

PeopleMakeCulture:
TODO: @PeopleMakeCulture to make PR to https://github.com/microbiomedata/NMDC_documentation/blob/main/docs/howto_guides/api_gui.md 

- [ ] add API version 
- [ ] remove documentation for old endpoints
- [ ] invite/request documentation on new endpoints (eg find/data_objects/study/study_id)

dwinston:
@PeopleMakeCulture #455 was merged. Can this issue be closed?

PeopleMakeCulture:
I'd like to keep it open as a reminder to monitor a couple loose ends (in checklist above)

PR #455, opened 2024-01-24
437 co-locate API usage documentation on SwaggerUI 

PeopleMakeCulture:
Pull end-user documentation from [How To Guides](https://github.com/microbiomedata/NMDC_documentation/blob/main/docs/howto_guides/api_gui.md) directly into Swagger UI

See: https://github.com/microbiomedata/nmdc-runtime/issues/437


dwinston (approved):
seems fine. will be good to have @brynnz22 look over this on api-dev.

brynnz22:
@dwinston can you link the dev api so I can look? I'm not sure how to access it.

brynnz22:
Thanks @dwinston . Its great how the documentation is incorporated. The only thing I noticed is there is a weird icon next to the `next_page_token` in the table for the metadata endpoints and also for the `next_cursor` for the find endpoints. I think its from the image in the doc I wrote that isn't translating. Can that be removed somehow from the GUI?

dwinston:
@brynnz22 is your last comment/question still applicable? I don't see any weird icon -- maybe I'm looking in the wrong place. Can you please attach a screenshot of the issue if it's still an issue?

brynnz22:
@dwinston yes, the icon is still there. Its in `Description` column of both for `page_token` row of the Metadata endpoint documentation drop down. And its in the `Description` column of the `cursor` row of the Find endpoint documentation dropdown. It looks like an image icon that is not being properly being pulled in.

PR #465, opened 2024-01-31
Fix inaccurate comment in migration instructions

eecavanna:
### Summary of changes

- Updated a statement that incorrectly said the origin server was typically local (not true) and the transformer server is typically remote (not true); to say the opposite.


eecavanna:
I'm OOO today. I wanted to submit this fix ASAP because other people may be using this Notebook soon and I thought they would be confused by the inaccuracy. I am ready for this branch to be merged into `main`.

pkalita-lbl (approved):

PR #500, opened 2024-03-26
fix: remove unused file

dwinston:
thanks @eecavanna

# Description

> Eric: I was surprised to see there is a nmdc.schema.json file in there. At a glance, it looks to me like a copy of the JSON Schema version of the NMDC schema, but it hasn't been updated (there) since it was first committed (some time last year). Is that being used by anything? Maybe it was used temporarily during early development and preserved in the repo by accident.

> Donny: re: 1 -- yes, this was preserved by accident. The schema is loaded from the installed package, e.g. https://github.com/microbiomedata/nmdc-runtime/blob/v1.4.2/nmdc_runtime/minter/config.py#L17 to source typecodes.

## Type of change

# How Has This Been Tested?

Inspection.

# Checklist:

n/a


eecavanna (approved):
Looks good to me. Thanks for removing this file.

PR #565, opened 2024-06-19
Configure GitHub Actions to push container image to GitHub Container Registry (in addition to Docker Hub)

eecavanna:
# Description

In this branch, I updated a GitHub Actions workflow so that, when it builds a container image and pushes it to Docker Hub, it also pushes a copy to GitHub Container Registry. That will make `nmdc-runtime` work more similarly to other NMDC products/applications (i.e. `nmdc-server`, `nmdc-edge`, and `nmdc-aggregator`).

Fixes #564 

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

It has not been tested yet. The final test will be whether Runtime package(s) appear at https://github.com/orgs/microbiomedata/packages once this new GHA workflow runs.

**Configuration Details**: none

# Checklist:

- [ ] My code follows the style guidelines of this project (have you run `black nmdc_runtime/`?)
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (in `docs/` and in <https://github.com/microbiomedata/NMDC_documentation/>?)
- [ ] I have added tests that prove my fix is effective or that my feature works, incl. considering downstream usage (e.g. <https://github.com/microbiomedata/notebook_hackathons>) if applicable.
- [ ] New and existing unit and functional tests pass locally with my changes (`make up-test && make test-run`)

PR #566, opened 2024-06-19
Authenticate as workflow triggerer instead of as repo owner

eecavanna:
This is an attempt to fix the HTTP 403 Forbidden error that occurred in this [GHA workflow run](https://github.com/microbiomedata/nmdc-runtime/actions/runs/9587845790/job/26438682126), which was triggered by the merging of PR https://github.com/microbiomedata/nmdc-runtime/pull/565.

```fortran
Error: buildx failed with: ERROR: failed to solve: failed to push ghcr.io/microbiomedata/nmdc-runtime-fastapi:main: failed commit on ref "manifest-sha256:***": unexpected status from PUT request to https://ghcr.io/v2/microbiomedata/nmdc-runtime-fastapi/manifests/sha256:***: 403 Forbidden
```

In this branch, I configured the GHA workflow step to authenticate with GHCR as the user triggering the workflow instead of as the repository owner. This makes it consistent with how it is done in [nmdc-server](https://github.com/microbiomedata/nmdc-server/blob/b0a10482a409167e2315c081706ecc35a8ba8816/.github/workflows/deploy.yml#L49) and [nmdc-edge](https://github.com/microbiomedata/nmdc-edge/blob/46139ee56ec05ae9898af5520522ce372c994725/.github/workflows/build-and-push-image.yml#L18) (it works in those repo)—instead of consistent with the [Docker documentation](https://docs.docker.com/build/ci/github-actions/push-multi-registries/) (which I had never tried before).

### DIFF
--- a/.github/workflows/build-and-release-to-spin.yml
+++ b/.github/workflows/build-and-release-to-spin.yml
@@ -65,11 +65,12 @@ jobs:
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
       # Reference: https://docs.docker.com/build/ci/github-actions/push-multi-registries/
+      # Reference: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
-          username: ${{ github.repository_owner }}
+          username: ${{ github.actor }}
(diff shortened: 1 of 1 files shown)


eecavanna:
Merging without review. This is an attempt to fix GitHub Actions and is not trivial for me to test by means other than merging into `main`.

PR #572, opened 2024-06-24
Update `NmdcPortalApiClient` with new authentication scheme

### LINKED ISSUE #571: Update `NmdcPortalApiClient` to use new Bearer token authentication
With https://github.com/microbiomedata/nmdc-server/pull/1257 the Portal API has a new authentication mechanism. The class in `nmdc-runtime` (`NmdcPortalApiClient`) which interacts with that API needs to be updated in accordance. Instead of storing a session cookie in the environment, we need to store a refresh token ([see also](https://github.com/microbiomedata/nmdc-server/issues/1277)). A `NmdcPortalApiClient` instance needs to exchange that refresh token for an access token and then use the access token in the `Authorization` header with any subsequent requests.

pkalita-lbl:
# Description

The Portal API has replaced session cookie-based authentication with Bearer token authentication. These changes update `NmdcPortalApiClient` to reflect that.

In testing I also found that the `SubmissionPortalTranslator` has drifted away from the Portal a bit. In particular, `samp_name` is now the unique key slot on `Biosample` instances instead of `source_mat_id`. These changes also correct that.

Fixes #571 

## Type of change

# How Has This Been Tested?

- [x] Unit test suite 
  - These changes update `tests/test_ops/test_data_api_ops.py` to reflect the new behavior of `NmdcPortalApiClient`.
  - These changes fix the test configuration of `tests/test_graphs/test_submission_portal_graphs.py` and remove the `xfail` mark.
- [x] Manually ran the `translate_metadata_submission_to_nmdc_schema_database` job in the Dagit interface

# Checklist:

- [ ] I have made corresponding changes to the documentation (in `docs/` and in <https://github.com/microbiomedata/NMDC_documentation/>?)
- [x] I have added tests that prove my fix is effective or that my feature works, incl. considering downstream usage (e.g. <https://github.com/microbiomedata/notebook_hackathons>) if applicable.
- [x] New and existing unit and functional tests pass locally with my changes (`make up-test && make test-run`)

### DIFF
--- a/nmdc_runtime/site/resources.py
+++ b/nmdc_runtime/site/resources.py
@@ -371,16 +371,31 @@ def gold_api_client_resource(context: InitResourceContext):
 
 @dataclass
 class NmdcPortalApiClient:
+
     base_url: str
-    # Using a cookie for authentication is not ideal and should be replaced
-    # when this API has an another authentication method
-    session_cookie: str
+    refresh_token: str
+    access_token: Optional[str] = None
+    access_token_expires_at: Optional[datetime] = None
+
+    def _request(self, method: str, endpoint: str, **kwargs):
+        if self.access_token is None or datetime.now() > self.access_token_expires_at:
+            refresh_response = requests.post(
+                f"{self.base_url}/auth/refresh",
+                json={"refresh_token": self.refresh_token},
+            )
+            refresh_response.raise_for_status()
+            refresh_body = refresh_response.json()
+            self.access_token_expires_at = datetime.now() + timedelta(
+                seconds=refresh_body["expires"]
+            )
+            self.access_token = refresh_body["access_token"]
+
+        headers = kwargs.get("headers", {})
+        headers["Authorization"] = f"Bearer {self.access_token}"
+        return requests.request(method, f"{self.base_url}{endpoint}", **kwargs, headers=headers)
 
     def fetch_metadata_submission(self, id: str) -> Dict[str, Any]:
-        response = requests.get(
-            f"{self.base_url}/api/metadata_submission/{id}",
-            cookies={"session": self.session_cookie},
-        )
+        response = self._request("GET", f"/api/metadata_submission/{id}")
         response.raise_for_status()
         return response.json()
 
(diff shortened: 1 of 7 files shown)


eecavanna (dismissed):

nmdc_runtime/site/resources.py
```diff
@@ -371,30 +371,47 @@ def gold_api_client_resource(context: InitResourceContext):
...
+    def _request(self, method: str, endpoint: str, **kwargs):
+        if self.access_token is None or datetime.now() > self.access_token_expires_at:
+            refresh_response = requests.post(
+                f"{self.base_url}/auth/refresh",
+                json={"refresh_token": self.refresh_token},
+            )
+            refresh_response.raise_for_status()
+            refresh_body = refresh_response.json()
+            self.access_token_expires_at = datetime.now() + timedelta(
+                seconds=refresh_body["expires"]
```
eecavanna, line 389: Side note (not the focus of this PR): I'm surprised that the Data Portal API returns the token's expiration as an offset, instead of as an absolute time. It seems to me that that could lead to there being an offset between the time the client thinks it expires and the time the server thinks it expires (e.g. a client on a slow internet connection might think it expires a couple seconds later than the server does).

pkalita-lbl: `nmdc-server` is _not_ a full OAuth 2.0 authorization server, but the `expires` field is inline with OAuth 2.0 recommendation of sending the lifetime in second: https://datatracker.ietf.org/doc/html/rfc6749#section-5.1. I don't know if using `expires` instead of `expires_in` was intentional or an oversight.

pkalita-lbl: I think it was an oversight so I'm requesting we change it to `expires_in`: https://github.com/microbiomedata/nmdc-server/pull/1287 Once that's accepted and merged I'll updated this PR as well.

eecavanna: Thanks for providing links to the recommendation. I agree that calling it `expires_in` (which is what it's called in the latest commit on this branch) and having it be a number of seconds (as opposed to an absolute timestamp) is in line with the recommendation.

eecavanna (approved):
Looks good to me!

PR #574, opened 2024-06-25
Update notebook to use revised migrator (via newer `nmdc-schema` PyPI package)

eecavanna:
Note: I erased the usual PR boilerplate content from this PR.

In this branch, I updated a migration notebook to use a newer version of the `nmdc-schema` PyPI package. Using a newer version was necessary in order to get access to a recently-revised migrator.

The revised migrator uses an additional collection, so I added that to the notebook.

Also, I fixed the format string for the logger (while I was already editing the notebook).


eecavanna:
Merging without review (currently the standard operating procedure for last-minute migrator changes).

Issue #600, opened 2024-07-18
`berkeley`: Fix failing tests in GOLD, NEON, and submission portal translators


eecavanna:
The test `test_get_pi` in `tests/test_data/test_gold_translator.py` is failing on the `berkeley` branch.

Here's an excerpt of the `pytest` [output](https://github.com/microbiomedata/nmdc-runtime/actions/runs/9983464838/job/27591102366) from the tip of the `berkeley` branch:

```py
demo/metadata_migration/notebooks/test_helpers.py .                      [  0%]
tests/e2e/test_minter_api.py XXXX                                        [  4%]
tests/integration/test_minter_repository.py ....X...                     [ 12%]
tests/test_api/test_db_mongo.py .....                                    [ 17%]
tests/test_api/test_endpoints.py ss.....s.                               [ 25%]
tests/test_api/test_metadata.py ssss..                                   [ 31%]
tests/test_data/test_gold_translator.py F
```

...

```py
tests/test_data/test_gold_translator.py:16: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
nmdc_runtime/site/translation/gold_translator.py:68: in _get_pi
    return nmdc.PersonValue(
<string>:10: in __init__
    ???
/usr/local/lib/python3.10/site-packages/nmdc_schema/nmdc.py:1058: in __post_init__
    super().__post_init__(**kwargs)
/usr/local/lib/python3.10/site-packages/nmdc_schema/nmdc.py:910: in __post_init__
    self.MissingRequiredField("type")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = PersonValue(type=None, has_raw_value='Joan D. Berger', email='jdb@example.com', name='Joan D. Berger', orcid=None, profile_image_url=None, websites=[])
field_name = 'type'

    def MissingRequiredField(self, field_name: str) -> None:
        """ Generic loader error handler """
>       raise ValueError(f"{field_name} must be supplied")
E       ValueError: type must be supplied

/usr/local/lib/python3.10/site-packages/linkml_runtime/utils/yamlutils.py:278: ValueError
```

...

```py
=========================== short test summary info ============================
[...]

PR #621, opened 2024-08-07
Fix Failing Tests in Berkeley Branch

aclum:
# Description

Changes from the Berkeley schema resulted in a number of test failures, mostly due to missing "type" fields, which are required in the Berkeley schema.

Three remaining tests are failing. These have been skipped for now in order to unblock the CI pipeline. They will be spun into separate GH issues. 

Fixes #  https://github.com/microbiomedata/nmdc-runtime/issues/600

## Type of change

# How Has This Been Tested?
Ran docker commands for testing (see Makefile)

### DIFF
--- a/nmdc_runtime/site/translation/submission_portal_translator.py
+++ b/nmdc_runtime/site/translation/submission_portal_translator.py
@@ -127,6 +127,7 @@ class SubmissionPortalTranslator(Translator):
             email=study_form.get("piEmail"),
             orcid=study_form.get("piOrcid"),
             profile_image_url=self.study_pi_image_url,
+            type=nmdc.PersonValue.class_class_curie,
         )
 
     def _get_doi(self, metadata_submission: JSON_OBJECT) -> Union[List[nmdc.Doi], None]:
--- a/nmdc_runtime/site/translation/submission_portal_translator.py
+++ b/nmdc_runtime/site/translation/submission_portal_translator.py
@@ -218,7 +218,10 @@ class SubmissionPortalTranslator(Translator):
         if not match:
             return None
 
-        qv = nmdc.QuantityValue(has_raw_value=raw_value)
+        qv = nmdc.QuantityValue(
+            has_raw_value=raw_value,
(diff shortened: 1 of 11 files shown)


aclum:
@sujaypatil96  not sure if this is useful or not. I updated some of the test assertions but the functions themselves need to be changed before tests will pass.

nmdc_runtime/site/translation/submission_portal_translator.py
```diff
@@ -127,6 +127,7 @@ def _get_pi(
             email=study_form.get("piEmail"),
             orcid=study_form.get("piOrcid"),
             profile_image_url=self.study_pi_image_url,
+            type=nmdc.PersonValue.class_class_curie,
```
aclum, line 130: @sierra-moxon @turbomam  what was your recommendation today instead of class_class_curie to return the type curie?

PeopleMakeCulture:
@aclum I couldn't tag you as a reviewer but it would be good to get your input too

nmdc_runtime/site/translation/submission_portal_translator.py
sierra-moxon: e.g. something like:
```
pv_details = self.schema_view.get_class("PersonValue")
pv_type = pv_details.class_uri
```

PeopleMakeCulture: > e.g. something like:
> 
> ```
> pv_details = self.schema_view.get_class("PersonValue")
> pv_type = pv_details.class_uri
> ```

Thanks @sierra-moxon that makes sense

PeopleMakeCulture:
> Ideally we'd adopt Sierra's suggest for specifying type in all the functions but @sujaypatil96 can update those time permitting. Otherwise this PR looks good to me.

Agreed!

dwinston (approved):
The show must go on. We will revisit the `@pytest.mark.xfail`-decorated tests at a later time. A difficult decision, but I agree with it given the situation.

PR #656, opened 2024-08-28
Update GOLD translator to work with Berkeley schema

sujaypatil96:
# Description

Updating the GOLD translator to work with the berkeley schema.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration, if it is not simply `make up-test && make test-run`.

- [ ] Updates to `tests/test_data/test_gold_translator.py`

# Definition of Done (DoD) Checklist:

- [ ] I have made corresponding changes to the documentation (in `docs/` and in <https://github.com/microbiomedata/NMDC_documentation/>?)
- [ ] I have added tests that prove my fix is effective or that my feature works, incl. considering downstream usage (e.g. <https://github.com/microbiomedata/notebook_hackathons>) if applicable.
- [ ] New and existing unit and functional tests pass locally with my changes (`make up-test && make test-run`)

### DIFF
--- a/nmdc_runtime/site/translation/gold_translator.py
+++ b/nmdc_runtime/site/translation/gold_translator.py
@@ -10,6 +10,7 @@ class GoldStudyTranslator(Translator):
     def __init__(
         self,
         study: JSON_OBJECT = {},
+        study_type: str = "research_study",
         biosamples: List[JSON_OBJECT] = [],
         projects: List[JSON_OBJECT] = [],
         analysis_projects: List[JSON_OBJECT] = [],
@@ -19,6 +20,7 @@ class GoldStudyTranslator(Translator):
         super().__init__(*args, **kwargs)
 
         self.study = study
+        self.study_type = nmdc.StudyCategoryEnum(study_type)
         self.biosamples = biosamples
         self.projects = projects
         self.analysis_projects = analysis_projects
@@ -69,7 +71,7 @@ class GoldStudyTranslator(Translator):
             has_raw_value=pi_dict.get("name"),
             name=pi_dict.get("name"),
             email=pi_dict.get("email"),
-            type=pi_dict.get("type"),
+            type="nmdc:PersonValue",
         )
 
     def _get_mod_date(self, gold_entity: JSON_OBJECT) -> Union[str, None]:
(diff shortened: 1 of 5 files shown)


sujaypatil96:
@aclum i had a question about populating the `instrument_used` slot on the `NucleotideSequencing` class.

Currently, we are getting the instrument information from `seqMethod` field from the GOLD API, and that information looks like

> Illumina HiSeq 2500-1TB

See https://gold.jgi.doe.gov/project?id=Gp0127656 for example.

Can we make the assumption that all `seqMethod` values will be formatted in the same way on GOLD's end? The reason I ask is because we need to create `Instrument` objects corresponding to each of those instruments and store them in the database.

aclum:
note for @turbomam  and @sujaypatil96  that in some cases the CV from JGI the model includes output info, for example NovaSeq S4 is instrument + flowcell version so we should pick the closest mapping. same for novaseq x see https://www.illumina.com/systems/sequencing-platforms/novaseq-x-plus/specifications.html

flowcells get loaded on instruments rather than being part of the instrument itself https://www.sciencedirect.com/topics/engineering/flow-cell#:~:text=An%20Illumina%20flow%20cell%20is,2).

nmdc_runtime/site/translation/gold_translator.py
```diff
@@ -19,9 +22,11 @@ def __init__(
         super().__init__(*args, **kwargs)
 
         self.study = study
+        self.study_type = nmdc.StudyCategoryEnum(study_type)
         self.biosamples = biosamples
         self.projects = projects
         self.analysis_projects = analysis_projects
+        self.gold_instrument_set_mapping_file_path = "https://raw.githubusercontent.com/microbiomedata/berkeley-schema-fy24/main/assets/misc/gold_seqMethod_to_nmdc_instrument_set.tsv"
```
pkalita-lbl, line 29: I always like to see these `Translator` classing being "pure", meaning there are no side effects like fetching external resources. To that end I would suggest passing in the actual mapping information as a constructor argument -- kind of like how you're passing in the `study_type` here, except the instrument mapping info would be a `list` or `dict` (either could work) instead of a string.

The `get_csv_rows_from_url` could be useful for doing the actual fetching and the URL being fetch should _definitely_ be part of the job config (again, similar to how you're already handling the `study_type`); we don't want to be going back in and updating this code when the `berkeley-schema-fy24` repo goes away. See existing uses of `get_csv_rows_from_url` for guidance.

pkalita-lbl (changes requested):

aclum:
@sujaypatil96  can you backmerge berkeley into this branch? There are failures for tests/test_api/test_endpoints.py which should be resolved in the berkeley branch

nmdc_runtime/site/translation/gold_translator.py
sujaypatil96: Great point! I've made changes to relevant files such that `GoldStudyTranslator` takes in the mapping info as a pandas DataFrame directly using an existing `@op` called `get_df_from_url` which handles TSV files, as opposed to `get_csv_rows_from_url` which works nicely for CSV files.

sujaypatil96:
@aclum it looks like certain Actions are erroring out at the tip of the `berkeley` branch even now? See the latest [commit](https://github.com/microbiomedata/nmdc-runtime/commit/66b36e994a8542034542f74f25c0fa02eafb44aa).

pkalita-lbl (approved):
Thanks for making those changes. Looks good to me now.

PR #664, opened 2024-08-31
Remove test code's dependence on a static Mongo dump

eecavanna:
Here are the changes I made in this PR branch:
1. **I removed the test set up code** that downloaded and restored a static Mongo dump that conformed to a specific version of `nmdc-schema`.
   1. That code was not compatible with branches of the Runtime on which a different version of `nmdc-schema` was in use. In practice, we saw test failures when the Runtime was using the "Berkeley schema" (i.e. `nmdc-schema` version 11.x); but I think we would have seen test failures even when the Runtime was using certain versions of the legacy schema (i.e. `nmdc-schema` version 10.x), depending upon how that schema version differed from the one the dump happens to conform to.
2. **I removed the test** that depended upon the aforementioned (removed) set up code loading specific data into the `alldocs` collection.
   1. **I replaced that test** with one that confirms that the endpoint (i.e. the endpoint being tested) returns an unsuccessful HTTP response when there is no `Study` having the specified `id`.
   3. I added a `TODO` comment about adding tests for the scenario where the `Study` _does_ exist (in which case, the endpoint would utilize the `alldocs` collection). This is what the original (removed) test did, but it depended upon the problematic set up code I referred to in item 1.
3. I also removed some unused `import` statements (I think they had been left there by mistake).

### DIFF
--- a/tests/test_api/test_endpoints.py
+++ b/tests/test_api/test_endpoints.py
@@ -33,70 +33,9 @@ def ensure_schema_collections_and_alldocs():
         )
         return
 
-    # Seed the database with documents that will be included in the `alldocs` collection.
-    #
-    # Note: An earlier version of this function seeded the database by restoring a static dump into it.
-    #       The dump was created from a database that conformed to a specific version of the "legacy"
-    #       schema (i.e. `nmdc-schema` v10.8.0). That schema version was not necessarily the same schema
-    #       version as the Runtime was using when this test was _run_.
-    #
-    #       The current version of the function contains a quick-and-dirty workaround designed to make the
-    #       function work with the latest versions of both the "legacy" schema and the "Berkeley" schema.
-    #
-    # TODO: Replace the workaround below with a solution that will accommodate schema changes over time
-    #       and that doesn't rely on a contrived collection.
-    #
-    # Docs: https://microbiomedata.github.io/nmdc-schema/ControlledIdentifiedTermValue/
-    # Docs: https://microbiomedata.github.io/berkeley-schema-fy24/ControlledIdentifiedTermValue/
-    controlled_identified_term_value = dict(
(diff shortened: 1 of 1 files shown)


tests/test_api/test_endpoints.py
```diff
@@ -60,8 +33,10 @@ def ensure_schema_collections_and_alldocs():
         )
         return
 
-    dump_dir = ensure_local_mongodump_exists()
-    mongorestore_from_dir(mdb, dump_dir, skip_collections=["functional_annotation_agg"])
+    # FIXME: Seed the database with documents that would be included in an `alldocs` collection,
```
dwinston, line 36: thank you for signposting via FIXME

tests/test_api/test_endpoints.py
```diff
@@ -374,10 +349,19 @@ def test_get_class_name_and_collection_names_by_doc_id():
...
-def test_find_data_objects_for_study(api_site_client):
+def test_find_data_objects_for_nonexistent_study(api_site_client):
+    r"""
+    Confirms the endpoint returns an unsuccessful status code when no `Study` having the specified `id` exists.
+    Reference: https://docs.pytest.org/en/stable/reference/reference.html#pytest.raises
+
+    Note: The `api_site_client` fixture's `request` method will raise an exception if the server responds with
+          an unsuccessful status code.
+
+    TODO: Add tests focused on the situation where the `Study` _does_ exist.
```
dwinston, line 360: 👍

dwinston (approved):
Yes, I'll re-think this to accommodate the more general case that the test mongodump should correspond (e.g. in its filepath) to a target schema version, and furthermore should e.g. fail informatively re: no corresponding mongodump if it is to fail at all in that case.

PR #697, opened 2024-09-20
Update field name to be consistent with Berkeley schema

### LINKED ISSUE #689: `berkeley`: API endpoint `/data_objects/study/{study_id}` not returning correct results
**Describe the bug**
This endpoint is returning an empty array b/c alldocs doesn't exist in berkeley

**To Reproduce**
1. curl -X 'GET' \
  'https://api-berkeley.microbiomedata.org/data_objects/study/nmdc%3Asty-11-33fbta56' \
  -H 'accept: application/json'
2. compare empty return body to results on production

**Expected behavior**
Response json body should contain the same records as production.

Example user story template:
AS A {user|persona|system},
[INSTEAD OF {current condition}]
I EXPECTED {result}
[SO THAT {value or justification}]
[NO LATER THAN {best by date}]

**Screenshots**

**Acceptance Criteria**
Response body  is ~51K in size. 

Example scenario-based template:
Given (some given context or precondition), when (I take this action), then (this will be the specific outcome).

**Additional context**
Consider adding test coverage for this endpoint.

cc @sujaypatil96 because ncbi export code also needs alldocs

eecavanna:
# Description

In this branch, I updated the name of a field used in a database query, to be consistent with the [Berkeley schema](https://microbiomedata.github.io/berkeley-schema-fy24/associated_studies/) (instead of the legacy schema). This fixes #689.

While I was there, I also updated the docstring to say that the function returns data objects as opposed to data object IDs.

## Type of change

# How Has This Been Tested?

- [ ] I'll let GHA test it

# Definition of Done (DoD) Checklist:

### DIFF
--- a/nmdc_runtime/api/endpoints/find.py
+++ b/nmdc_runtime/api/endpoints/find.py
@@ -139,21 +139,25 @@ def find_data_objects_for_study(
     study_id: str,
     mdb: MongoDatabase = Depends(get_mongo_db),
 ):
-    """This API endpoint is used to retrieve data object ids associated with
-    all the biosamples that are part of a given study. This endpoint makes
+    """This API endpoint is used to retrieve data objects associated with
+    all the biosamples associated with a given study. This endpoint makes
     use of the `alldocs` collection for its implementation.
 
     :param study_id: NMDC study id for which data objects are to be retrieved
     :param mdb: PyMongo connection, defaults to Depends(get_mongo_db)
-    :return: List of dictionaries where each dictionary contains biosample id as key,
-        and another dictionary with key 'data_object_set' containing list of data object ids as value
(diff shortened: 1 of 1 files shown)


eecavanna:
Merging without review (in the interest of time) before I refresh the Berkeley environment's Mongo database. These changes have no impact on the results of automated tests, according to GHA. These changes impact only a single endpoint, which is currently not working correctly (this PR contains a bugfix).

sujaypatil96 (commented):
Looks good! you're right, `part_of` does not exist anymore in berkeley schema, instead we need to use `associated_studies` like you are 🚀

eecavanna:
Thank you for reviewing the changes, @sujaypatil96!

PR #709, opened 2024-09-28
Fix bug where one document in each collection is omitted from `alldocs`

eecavanna:
In this branch, I fixed a bug in the function that generates the `alldocs` collection. The bug was causing the `alldocs` collection to lack one document from each upstream collection.

This fixes the `E` failures that pytest recently began reporting on the `berkeley` branch (their absence is shown here).

image


eecavanna:
Merging without review, based on the urgency and the fact that we've had discussions related to this (albeit at a higher level) on Slack within the past couple hours, and the fact that the target branch is not `main`.

Issue #434, opened 2024-01-10
API endpoint `/queries:run` only returns the first 100 matching documents


mbthornton-lbl:
queries:run endpoint truncates data in response
The queries:run endpoint is only returning the first 100 results of a query

Query:
```json
{
  "find": "data_generation_set",
  "filter": {"associated_studies": "nmdc:sty-11-aygzgv51"}
}
```
Response:
[response_1704828951248.json](https://github.com/microbiomedata/nmdc-runtime/files/13891193/response_1704828951248.json)

Note at the end of the response JSON that it claims partialResultsReturned: null

aclum:
this needs prioritization @shreddd

aclum:
@PeopleMakeCulture @dwinston  please comment on the effort needed to add paging.

PeopleMakeCulture:
TODO:
Determine what is the mean document size for collections? If the docs are small enough, we can increase the default size.
Previous to Berkeley schema, proteomics has some docs that were up to several MB large, which is why we limited API requests by default to 20 docs. If this is no longer the case we can easily increase the default doc limit to return

dwinston:
TODO: update collection name from "omics_processing_set" in @mbthornton-lbl 's relevant query, to ensure reproduction for solving this issue.

aclum:
I updated the description in the example to be valid for berkeley.

Documents in workflow_exeuction_set still are quite large.

PeopleMakeCulture:
We are currently exploring two approaches to get this done w/ limited time:
1. Leverage and document mongo’s skip & limit query syntax
2. Re-use cursor implementation from other `find` endpoints

PeopleMakeCulture:
@mbthornton-lbl @aclum The above query can be executed using the `/nmdcschema/{collections}` endpoint like so. The limit can likely be increased past 100. The response will contain a token at the bottom which can be passed in the next API call to get the next set of documents. Is this sufficient for your use case?

![Image](https://github.com/user-attachments/assets/329bca52-d0ac-4f84-b788-b4af674a2ee5)

aclum:
We are not currently blocked so don't need a workaround, the request here is for the queries:run endpoint to be able to return a complete set of records. queries:run is the only endpoint right now with the flexibility to traverse collections with a query beside study id and where you have any control over the information returned.

aclum:
[...]

PR #754, opened 2024-11-06
Add pagination to `/queries:run` API endpoint (`find` and `aggregate`)

dwinston:
On this branch, we added cursor-based pagination to the `/queries:run` API endpoint.

#### Pagination for `find` and `aggregate` commands

We implemented pagination for the `find` and `aggregate` commands.

When the server receives one of these commands, it queries the database and sorts the results by their `_id` values. It then returns the first batch of results (the batch size is configurable via the API request and there is also a default batch size).

If the number of results equals or exceeds the batch size (note: the server can't tell the difference), the server will includes a pagination token (i.e. some string) in the `cursor.id` field of the response. Clients can use that pagination token with a `getMore` command to get the next batch (see next bullet point).

#### Pagination token storage on the server side

The original client-defined command and the last `_id` value from the most recently-retrieved batch are stored in the MongoDB database. Those documents have a finite TTL, ensuring they are automatically deleted (by Mongo, itself) after a certain amount of time passes (this will prevent them from building up in our database).

When the server receives a `getMore` command (which will include a pagination token), the server uses the received pagination token to find the original command and the last `_id` value from the database. It then uses that original command and last `_id` value, to fetch and return the next batch of results (along with — if this batch is also full — _another_ pagination token).

#### Documentation

We updated the endpoint's Swagger UI documentation to include details about pagination, including its limitations. We also refined the existing example payloads.

### Related issue(s)

Fixes https://github.com/microbiomedata/nmdc-runtime/issues/434

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

We tested these changes by adding automated tests and confirming they all pass, both locally and on GHA. We also tested the new functionality with a recent production dump and a complex query provided by a team member.

### Documentation

- [ ] Other (explain below)

### Maintainability


dwinston:
I realized that I could implement this in a way compatible with the current `POST /queries:run` API (though we still may choose to deploy this functionality as a new API endpoint), by taking liberty with some of the currently-unusable mongo command interfaces wrt `find` and `aggregate` command cursors and `getMore` commands. Specifically, they (cursor-aware commands) haven't been usable via the API because they depend on MongoDB (server) sessions that [are specified](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.md) to not be usable across MongoClient instances and thus across HTTP requests.

What I've started, is to adapt the existing Mongo command API to pretend there's a backing server session, but really we're doing it with "ephemeral" (managed by a Mongo TTL index) database state (via an internal `_runtime.cursor_continuations` collection), analogous to how MongoDB manages sessions internally via its `system.sessions` collection.

.env.test
```diff
@@ -20,4 +20,7 @@ API_SITE_ID=nmdc-runtime
 API_SITE_CLIENT_ID=chws-tk74-51
 API_SITE_CLIENT_SECRET=070d56033559e3096d4f82d30aabdd51c3b8a57ac20d24f743a27444d5edc3db
 
-MINTING_SERVICE_ID=nmdc:minter_service_11
\ No newline at end of file
+MINTING_SERVICE_ID=nmdc:minter_service_11
+
+# Can append e.g. '/nmdc/biosample_set.bson.gz' to fetch the 'biosample_set' collection from the 'nmdc' database.
+MONGO_REMOTE_DUMP_URL_PREFIX=https://nmdc-runtime.files.polyneme.xyz/nmdc-mongodumps/dump_nmdc-prod_2025-02-12_20-12-02
```
dwinston, line 26: @shreddd @eecavanna here's the environment variable I referred to during the recent NMDC infrastructure sync, to point to a mirror of a subset of e.g. the `dump_nmdc-prod_2025-02-12_20-12-02` directory within NMDC's project directory on the NERSC filesystem.

dwinston:
Unfortunately, I won't be able to finish this off today (i.e. present a non-draft, merge-able PR) for inclusion in next week's (2025.02) release.

/cc @aclum @shreddd

eecavanna:
To anyone that has this branch checked out locally: I'm going to start ramping up on this WIP PR now. I will be reading through the diff _in edit mode_ and adding comments where I think they would facilitate development/maintenance. If I do end up adding comments, I'll commit them to the branch.

eecavanna:
I'll fix those merge conflicts.

eecavanna:
Hi @shreddd, I finished merging `main` into this PR branch.

eecavanna:
There are some merge conflicts on this branch, involving the `docker-compose.test.yml` and `docker-compose.yml` files. I assume the changes I introduced via https://github.com/microbiomedata/nmdc-runtime/pull/884 (merged into `main` 4 days ago) are creating the conflict. I'll resolve the conflicts on this branch after checking with @shreddd via Slack.

eecavanna:
OK, I finished resolving the merge conflicts.

Separately, @shreddd pointed out that the tests seem to take longer to run on this branch than on `main`. I am wondering whether running the tests on this branch involves downloading dumps from the Internet and restoring them into the test database. I believe that was the case a few weeks ago. That could be a/the cause of the longer test duration.

shreddd:
@eecavanna - the latest merge seems to be breaking my local test env because of how mongo is set up.

eecavanna:
I'm out right now. In an attempt to unblock you, I wanted to recommend you modify the mongo container image version to match whatever the data you have in your local database was created with (either 6.0.4 or 8.0.4).

eecavanna:
I'd like to pair debug that with you during tomorrow morning's referential integrity squad meeting. For anyone else reading this; in recent occurrences of that meeting, this PR is a topic that we have discussed.

eecavanna:
Before this branch gets merged into `main`, I want to update the `mongo` image on this branch from 6.0.4 to 8.0.4. It was temporarily downgraded to 6.0.4 to accommodate either (a) people's development environments, where they had Mongo 6 dumps loaded into their Mongo containers; (b) the production partial dump that the Runtime now fetches from the Internet when it runs its tests; or (c) both of the above.

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -47,18 +49,25 @@ def check_can_update_and_delete(user: User):
...
     """
-    Allows `find`, `aggregate`, `update`, and `delete` commands for users with permissions.
+    Allows `find`, `aggregate`, `update`, `delete`, etc. commands for users with permissions.
 
-    For `find` and `aggregate`, note that cursor batching/pagination does *not*
-    work via this API, so ensure that you construct a command that will return
-    what you need in the "first batch". Also, the maximum size of the returned payload is 16MB.
+    For `find` and `aggregate`, a `cursor` `id` may be returned, which can be used with a `getMore` command
+    to retrieve the next `batch` of documents. Note that the maximum size of the response payload is 16MB, so
+    you may need to set the `batchSize` parameter appropriately for `find` or `aggregate`.
```
eecavanna, line 70:
```suggestion
    For `find` and `aggregate` commands, the requested items will be in `cursor.batch`.
    Whenever there _may_ be more items available, the response will include a non-null `cursor.id`.
    To retrieve the next batch of items, submit a request with `getMore` set to that `cursor.id` value.
    When there are no more items available, the response will include a null `cursor.id`.

    Note that the maximum size of the response payload is 16MB, so
    you may need to set the `batchSize` parameter appropriately for `find` or `aggregate`.
```

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -246,6 +293,11 @@ def _run_mdb_cmd(cmd: Cmd, mdb: MongoDatabase = _mdb) -> CommandResponse:
             )
 
             cmd = AggregateCommand(**modified_cmd_doc)
+        else:
+            raise HTTPException(
+                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                detail="Invalid cursor continuation. Could not determine originating query.",
+            )
```
eecavanna, line 301: Oops! I didn't mean to include this in the commit. I added this `else` block locally as a reminder to myself that the "else" case was not being addressed explicitly in the code yet.

eecavanna:
It occurred to me this afternoon that... I don't think the Runtime validates that the user of a given `getMore` string is the same user that submitted the original query. I don't see any properties in the `QueryContinuation` class that keep track of the user that "created" it. Maybe this is not an issue right now, given the types of operations that can be "continued."

nmdc_runtime/api/models/query_continuation.py
```diff
@@ -0,0 +1,112 @@
...
+    QueryCmd,
+    CursorYieldingCommand,
+)
+
+_mdb: MongoDatabase = get_mongo_db()
+
+# Ensure one-hour TTL on `_runtime.query_continuations` documents via TTL Index.
+_mdb["_runtime.query_continuations"].create_index(
+    {"last_modified": 1}, expireAfterSeconds=3600
+)
```
eecavanna, line 37: I didn't know [TTL Indexes](https://www.mongodb.com/docs/manual/core/index-ttl/) were a thing.

> TTL indexes are special single-field indexes that MongoDB can use to automatically remove documents from a collection after a certain amount of time or at a specific clock time. Data expiration is useful for certain types of information like machine generated event data, logs, and session information that only need to persist in a database for a finite amount of time.

Cool! @dwinston

tests/test_api/test_endpoints.py
```diff
@@ -964,10 +963,39 @@ def test_run_query_aggregate_with_continuation(api_user_client):
                     }
                 },
             ],
+            # TODO: Why is this "lever" (whether the MongoDB server can use a disk) being exposed to the API client?
+            #       Also, according to the MongoDB docs, this is effectively `True` by default:
+            #       > "[...] pipeline stages that require more than 100 megabytes of memory to execute
+            #       > write temporary files to disk by default."
+            #       Reference: https://www.mongodb.com/docs/manual/reference/method/cursor.allowDiskUse
+            #
```
dwinston, line 972: Excellent. Yes, this is enabled by default now (c.f. <https://www.mongodb.com/docs/manual/reference/parameters/#mongodb-parameter-param.allowDiskUseByDefault>. So, we can remove this user-settable parameter from our interface.

dwinston: @eecavanna the "why" is that I've been using mongodb since version 3, and I don't always catch the improvements in default values. 😄

dwinston:
> It occurred to me this afternoon that... I don't think the Runtime validates that the user of a given `getMore` string is the same user that submitted the original query. I don't see any properties in the `QueryContinuation` class that keep track of the user that "created" it. Maybe this is not an issue right now, given the types of operations that can be "continued."

In previous commits, a user was tracked. As part of simplifying the model, I decided that user info shouldn't be an issue given that the `getMore` cursor id required has high entropy, i.e. it is effectively a "key" that a user is unlikely to be able to produce unless they were given it in a response body.

README.md
```diff
@@ -134,6 +134,9 @@ Tests can be found in `tests` and are run with the following commands:
 ```bash
 make up-test
 make test
+
+# Run a Specific test file eg. tests/test_api/test_endpoints.py
+make test ARGS="tests/test_api/test_endpoints.py"
```
dwinston, line 139: TIL! thanks @shreddd.

Just ran 
```
make test-run ARGS="-k 'test_run_query_aggregate_with_continuation'"
```

eecavanna:
I'm working on removing the test suite's dependency on an externally-hosted database dump.

When I comment out the `ensure_schema_collections_loaded()` statement within the `ensure_schema_collections_and_alldocs` function (in `test_endoints.py`), delete the `nmdc` database in the `mongo` container in the test stack, and re-run the tests via `make test-run`, I see the following three errors in the pytest output:

```console
====================================================== short test summary info =======================================================
FAILED tests/test_api/test_endpoints.py::test_run_query_find_with_continuation - assert None is not None
FAILED tests/test_api/test_endpoints.py::test_run_query_aggregate_with_continuation - assert 0 == 5
FAILED tests/test_ops/test_ops.py::test_apply_metadata_in_functional_annotation_agg - AttributeError: 'NoneType' object has no attribute 'id'
======================= 3 failed, 124 passed, 5 skipped, 6 xfailed, 1 xpassed, 4 warnings in 65.67s (0:01:05) ========================
```

I will focus on making those tests no longer depend upon a production dump. My plan at this point is to have the test code create documents relevant to the test, insert them into the database, run the test, then delete them from the database.

eecavanna:
I'm still working on that. It's taking me a while as there are existing tests that depend upon other tests leaving documents in the test database (i.e. tests are coupled to one another).

eecavanna:
In other words, I have implemented it for the pagination-related tests, but by deleting documents from the database in those tests, I have broken some changesheet-related tests.

eecavanna:
In that latest commit, I disabled the code that fetches the production dump from the Internet and updated the pagination-related tests to no longer rely on it (instead, they create documents relevant to the specific test at hand, and insert those into the database, and then delete them at the end of the test).

Disabling the code that fetches the production dump broke a test in `test_ops.py`. In addition, emptying out the `study_set` and/or `biosample_set` collection in the pagination tests broke two tests in `test_metadata.py`.

eecavanna:
The tests pass (on GHA—but not on my local machine for some reason) now, and they no longer depend upon a production dump. The code that would fetch the production dump is still in place, but the statement that invokes it has been removed. Everything else that was outstanding on Friday afternoon remains outstanding at this point (e.g. adding tests targeting corner cases).

eecavanna:
A note in case any of y'all come across the following statement in the Mongo docs. I don't think it impacts what we're working on in this PR.

I was reading the [Mongo docs](https://www.mongodb.com/docs/manual/reference/method/cursor.skip/#using-range-queries) just now and came across this statement:

> While [ObjectId](https://www.mongodb.com/docs/manual/reference/bson-types/#std-label-objectid) values should increase over time, they are not necessarily monotonic. This is because they:
> - Only contain one second of temporal resolution, so [ObjectId](https://www.mongodb.com/docs/manual/reference/bson-types/#std-label-objectid) values created within the same second do not have a guaranteed ordering...

image

I **don't** interpret this to mean that sorting the same set of documents by their `_id` values multiple times can put them in different orders. I think they'll always be sorted into the same order.

I **do** interpret it to mean that sorting documents by their `_id` values is not necessarily the same as sorting them by their creation times.

eecavanna:
The endpoint returns an HTTP 500 when the documents produced by the final stage of an aggregation lack an `_id` field. This is one of the scenarios listed in the TODO list created by @shreddd last week. The first "workaround" that comes to my mind is to have the endpoint check for that case and then change its behavior accordingly.

tests/test_api/test_endpoints.py
```diff
@@ -842,3 +937,336 @@ def test_run_query_update(api_user_client):
...
+    studies = generate_studies(6, title=study_title)
+    study_set.insert_many(studies)
+
+    # Fetch the first batch and confirm the `cursor.id` value is null, since
+    # pagination is not implemented when any of the documents lacks an `_id`
+    # field.
+    #
+    # TODO: Consider whether an API client might construct an aggregation
+    #       pipeline in which the output documents have custom `_id` values,
+    #       as opposed to the default `_id` values that MongoDB generates.
```
eecavanna, line 1107: This is an attack vector we have not discussed yet.

tests/test_api/test_endpoints.py
```diff
@@ -842,3 +937,336 @@ def test_run_query_update(api_user_client):
...
+    #
+    # FIXME: The endpoint-under-test does not handle this scenario yet, so we do
+    #        not assert anything about its response yet. Update the endpoint to
+    #        handle this scenario gracefully; then update this test to assert
+    #        things about the endpoint's response.
```
eecavanna, line 1112: When the documents output by the final stage of an aggregation pipeline lack an `_id` field, the endpoint "crashes" and the client gets an HTTP 500 response.

eecavanna:
On this branch, I updated the `mongo` service (in the dev/test stacks) to version 8.0.4. Here's how you can **delete** the `mongo` service's volume—which may contain data that is incompatible with Mongo 8—in your local dev/test stack:

```shell
# Delete the volume in the dev stack:
docker compose down --volumes mongo

# Delete the volume in the test stack:
docker compose -f docker-compose.test.yml down --volumes mongo
```

CC: @shreddd, @dwinston

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -219,24 +274,137 @@ def _run_query(query, mdb) -> CommandResponse:
...
+        )
+        if "find" in initial_cmd_doc:
+            modified_cmd_doc = assoc_in(
+                initial_cmd_doc,
+                ["filter", "_id", "$gt"],
+                cc.last_doc__id_for_cc(cursor_continuation),
+            )
+            cmd = FindCommand(**modified_cmd_doc)
+        elif "aggregate" in initial_cmd_doc:
+            # TODO assign `query` cmd to equivalent aggregate cmd.
```
eecavanna, line 298: I don't understand this comment.

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -219,24 +274,137 @@ def _run_query(query, mdb) -> CommandResponse:
...
+    )
+    if isinstance(cmd, CursorYieldingCommand):
+        batch_key = "firstBatch" if isinstance(cmd, QueryCmd) else "nextBatch"
+        cmd_response_adapted = assoc_in(
+            cmd_response_raw,
+            ["cursor", "batch"],
+            cmd_response_raw["cursor"][batch_key],
+        )
+        # TODO: Did you mean to delete this from `cmd_response_adapted` instead?
+        del cmd_response_raw["cursor"][batch_key]
```
eecavanna, line 346: Highlighting this `TODO` comment for reviewers.

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -219,24 +274,137 @@ def _run_query(query, mdb) -> CommandResponse:
...
+                    f"{'update' if isinstance(cmd, UpdateCommand) else 'delete'} command modified zero documents."
                     " I'm guessing that's not what you expected. Check the syntax of your request."
                     " But what do I know? I'm just a teapot.",
                 ),
             )
-    mdb.query_runs.insert_one(query_run.model_dump(exclude_unset=True))
+
+    # Cursor-command response? Prep runtime-managed cursor id and replace mongo session cursor id in response.
+    cursor_continuation = None
+    # TODO: Handle empty cursor response or situations where batch < batchSize.
```
eecavanna, line 372: Highlighting this `TODO` for reviewers.

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -219,24 +274,137 @@ def _run_query(query, mdb) -> CommandResponse:
...
+        # Append query run to current continuation
+        cursor_continuation = cc.get_cc_by_id(cursor_id)
+        cursor_continuation.cmd_responses.append(
+            CursorYieldingCommandResponse.cursor_batch__ids_only(cmd_response)
+        )
+        cmd_response.cursor.id = (
+            None if cmd_response.cursor.id == "0" else cursor_continuation.id
+        )
+        # TODO: remove print
+        print(f"getmore:{cmd_response.cursor.id=}")
```
eecavanna, line 409: Consider using a `logging` method instead.

tests/test_api/test_endpoints.py
```diff
@@ -842,3 +814,798 @@ def test_run_query_update(api_user_client):
...
+    #
+    # FIXME: Update tests that leave residue in the database to not do so, and
+    #        update the "dependent" tests accordingly. Or, (preferrably)
+    #        implement a database test fixture that always provides a "clean"
+    #        (i.e. re-initialized) database to each test that uses the fixture.
+    #        This could be scoped to the test function or to the test module,
+    #        the latter being an option for people that _do_ want specific
+    #        tests to be coupled to one another. This test fixture approach
+    #        would also accommodate tests that abort prematurely due to
+    #        failure and are, as a result, unable to run their "clean up" code.
```
eecavanna, line 850: @eecavanna — file a ticket about this if one does not already exist.

nmdc_runtime/api/endpoints/queries.py
eecavanna: I have resolved this locally and will push up the commit shortly.

nmdc_runtime/api/endpoints/queries.py
eecavanna: Locally, I have added some commentary below this `TODO`.

```py
    # TODO: Handle empty cursor response or situations where batch < batchSize.
    #
    #       Note: This "TODO" comment has not been removed, but — based upon the
    #             results of the automated tests, which do submit "find" and
    #             "aggregation" commands that produce empty result sets and result
    #             sets smaller than one batch — I think this has been resolved.
    #
```

CC: @shreddd

nmdc_runtime/api/endpoints/queries.py
eecavanna: Hi @dwinston, did you mean to delete a property of `cmd_response_raw` here (`cmd_response_raw` is not referenced after this line) or a property of `cmd_response_adapted` here (`cmd_response_adapted` is referenced a few lines later)?

nmdc_runtime/api/endpoints/queries.py
eecavanna: Hi @dwinston, the "git blame-y" feature of my text editor is saying this comment was added by you a few weeks ago. Given how the endpoint has developed since then, is this still an outstanding task?

nmdc_runtime/api/models/query_continuation.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def get_cc_by_id(cc_id: str) -> QueryContinuation | None:
+    doc = _coll_cc.find_one({"_id": cc_id})
+    if doc is None:
+        raise QueryContinuationError(f"cannot find cc with id {cc_id}")
+    return QueryContinuation(**doc)
+
+
+def last_doc__id_for_cc(cursor_continuation: QueryContinuation) -> str:
```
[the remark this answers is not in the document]
eecavanna: Hi @dwinston, I think you came up with these names pretty early on in the prototyping/development of the pagination solution. I'll defer to you about whether these are the same thing, conceptually.

eecavanna:
I finished updating the PR description.

I'd like to merge this in by Tuesday at 12pm PT. Then, it will become live on the "api-dev" site and team members (e.g. @aclum) will be able to experiment with it in preparation for its inclusion in next week's NMDC Release.

dwinston:
> On this branch, I updated the `mongo` service (in the dev/test stacks) to version 8.0.4. Here's how you can **delete** the `mongo` service's volume—which may contain data that is incompatible with Mongo 8—in your local dev/test stack:
> 
> ```shell
> # Delete the volume in the dev stack:
> docker compose down --volumes mongo
> 
> # Delete the volume in the test stack:
> docker compose -f docker-compose.test.yml down --volumes mongo
> ```

Thanks. Note also that `make down-test` deletes all volumes in the test stack, as that stack is intended to be particularly ephemeral, even locally.

nmdc_runtime/api/models/query_continuation.py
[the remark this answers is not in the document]
dwinston: My apologies: I had intended -- and still intend -- a total refactor/renaming from "cursor continuation" to "query continuation". I had used the former term at first.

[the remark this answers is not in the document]
dwinston: please go ahead and standardize to "query continuation" (and I suppose "qc").

nmdc_runtime/api/endpoints/queries.py
dwinston: Nope. Thanks.

nmdc_runtime/api/models/query_continuation.py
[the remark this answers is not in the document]
eecavanna: All good — thanks, @dwinston! I will make these name standardization changes now. It's the last change I plan to make on this branch before merging it into `main`.

eecavanna:
I'll merge this into `main` now without an official review on GitHub, in accordance with conversations that happened outside of GitHub (e.g. via Zoom and on Slack).

PR #779, opened 2024-11-18
Implement an API endpoint an admin can use to update a user's password

### LINKED ISSUE #778: Implement password reset flow for API users
**The Pain**

Currently, API users without administrative privileges cannot change/reset/update their passwords (for use in password-based auth). For an administrator to reset a user's password, they currently

1. create a new temporary user, issue a direct database command to set the value of the requesting user's hashed password to that of the temporary user, and delete the temporary user (as @eecavanna reported doing in https://github.com/microbiomedata/nmdc-runtime/issues/311#issuecomment-1936729900), or
2. delete and recreate the user with the same username, etc.

**The Dream**

As an API user,
instead of asking an admin to manually provide me with a new password,
I want to reset my password through email/orcid verification by myself so that I have a new password to use (and securely save) within a few minutes.

**The Fix (i.e., Acceptance Criteria)**
- [ ] Given that I remember and still have access to the email address associated with my API user record, when I call a reset-password endpoint with my email address as a request parameter, then I'll get a one-time magic link via email from which I can render and save a new, secure password.
- [ ] Given that I remember and still have access to the ORCiD account associated with my API user record, when I call a reset-password endpoint while authenticated via the ORCiD auth flow, then I'll get a one-time magic link via JSON response from which I can render and save a new, secure password.

**Alternatives**
Status quo. `<joke>`How often do users lose their passwords, anyway?`</joke>`

**Additional context**
Private communication with @shreddd via NMDC Slack raising this issue.

eecavanna:
In this branch, @shreddd implemented an API endpoint that administrators can use to update the password of an existing user.

### Details

(TODO)

### Related issue(s)

Fixes #778

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by implementing a `test_update_user` test function and ensuring that it passes.

### Documentation

- [x] Other (explain below)

This PR is being created by someone other than the branch author. I do not know the extent to which the branch author has considered documentation.

### Maintainability

This PR is being created by someone other than the branch author. I do not know which of the above things the branch author has done.


nmdc_runtime/api/endpoints/users.py
```diff
@@ -170,3 +182,32 @@ def create_user(
...
+    requester: User = Depends(get_current_active_user),
+    mdb: pymongo.database.Database = Depends(get_mongo_db),
+):
+    check_can_create_user(requester)
+    username = user_in.username
+
+    if user_in.password:
+        user_dict = UserInDB(
+            **user_in.model_dump(),
+            hashed_password=get_password_hash(user_in.password),
```
eecavanna, line 205: I think I will find this comment helpful on future read-throughs.

```suggestion
            # Store the password hash instead of the plain text password.
            hashed_password=get_password_hash(user_in.password),
```

tests/test_api/test_endpoints.py
```diff
@@ -173,6 +174,71 @@ def get_token():
...
+    base_url = os.getenv("API_HOST")
+
+    @retry(wait=wait_random_exponential(multiplier=1, max=60))
+    def get_token():
+        """
+
+        Randomly wait up to 2^x * 1 seconds between each retry until the range reaches 60
+        seconds, then randomly up to 60 seconds afterwards
+
+        """
```
eecavanna, line 189: I added a summary of what the function does, in addition to the retry aspect.

```suggestion
        """
        Fetch an auth token from the Runtime API, retrying until successful.
        
        Randomly wait up to 2^x * 1 seconds between each retry until the range reaches 60
        seconds, then randomly up to 60 seconds afterwards.
        Reference: https://pypi.org/project/retry/
        """
```

tests/test_api/test_endpoints.py
```diff
@@ -173,6 +174,71 @@ def get_token():
         )
 
 
+def test_update_user():
+    mdb = get_mongo(run_config_frozen__normal_env).db
+    rs = ensure_test_resources(mdb)
+    base_url = os.getenv("API_HOST")
+
+    @retry(wait=wait_random_exponential(multiplier=1, max=60))
```
eecavanna, line 182: FYI: The [docs](https://pypi.org/project/retry/) of the `retry` decorator say:

> ```
> :param tries: the maximum number of attempts. default: -1 (infinite).
> ```

I wonder whether this can lead to tests hanging indefinitely. I think that is what @shreddd was alluding to [here](https://github.com/microbiomedata/nmdc-runtime/issues/778#issuecomment-2480247793).

tests/test_api/test_endpoints.py
```diff
@@ -173,6 +174,71 @@ def get_token():
...
+        token_response = _rv.json()
+        return token_response["access_token"]
+
+    headers = {"Authorization": f"Bearer {get_token()}"}
+
+    user_in1 = UserIn(username="foo", password="oldpass")
+    mdb.users.delete_one({"username": user_in1.username})
+    mdb.users.update_one(
+        {"username": rs["user"]["username"]},
+        {"$addToSet": {"site_admin": "nmdc-runtime-useradmin"}},
```
eecavanna, line 208: I _think_ this test has an admin user updating its _own_ password. While that _is_ a valid use case, I am more interested in confirming that an admin user can update a _different_ user's password. I may be mistaken about what is happening in this test.

tests/test_api/test_endpoints.py
shreddd: I was wondering if we need the retry in the first place, but in talking to @dwinston this may be needed to account for async code. The tests shouldn't hang infinitely because it has a 60 second timeout.

eecavanna: I think the timeout (which can be up to 60 seconds long) can happen an infinite number of times (because the default value of `tries` is `-1`). That is where I got the idea of hanging indefinitely from.

eecavanna: In other words, the duration of each thing is limited; but the quantity of things is currently unlimited.

shreddd: Ah I see. We can add `stop=stop_after_attempt(N))` as an argument to retry. What is a good value for this N? Maybe 3 or 5?

shreddd: OK - added stop=stop_after_attempt(3)

eecavanna: Thanks, @shreddd. Stopping after three attempts sounds good to me.

Turns out I was looking at the documentation for the wrong PyPI package. Now that I've looked at the `import` statements in this Python file (I'm sorry I didn't do that earlier), I see that the package being used is `tenacity` (its documentation is [here](https://tenacity.readthedocs.io/en/stable/api.html)). I had incorrectly assumed that it was the older library called `retry` (which is documented [here](https://github.com/invl/retry)). That's why I kept referring to a kwarg named `tries` (instead of `stop`). By coincidence, they both do default to an infinite number of attempts (whether framed as "try infinity times" or "never stop trying").

dwinston (approved):
LGTM

eecavanna:
This LGTM also. I'll merge it in, which will cause it to be deployed to the development environment.

eecavanna:
By the way, @shreddd is aware that I was going to convert it from "Draft" to "Ready" and then merge it in (we talked a couple minutes ago).

PR #784, opened 2024-11-20
Delete unused `tests2` directory

shreddd:
This PR drops a directory that is not being used 
/tests2/


eecavanna (approved):
I'm OK with this branch being merged into `main`.

Git says this directory was created by Elais back in 2022 (via [this commit](https://github.com/microbiomedata/nmdc-runtime/commit/dcb77b267259ff0b287d95de8785779811d43091)). It has not been updated since then.

PR #788, opened 2024-11-21
Replace deprecated `utcnow()` calls with `now(timezone.utc)`

shreddd:
Fix to address deprecated datetime.utcnow() calls

### DIFF
--- a/demo/metadata_migration/notebooks/bookkeeper.py
+++ b/demo/metadata_migration/notebooks/bookkeeper.py
@@ -1,6 +1,6 @@
 from typing import Optional
 from enum import Enum
-from datetime import datetime
+from datetime import datetime, timezone
 
 from pymongo import MongoClient
 from nmdc_schema.migrators.migrator_base import MigratorBase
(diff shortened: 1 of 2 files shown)


eecavanna (approved):
I see the deprecation notice in the [Python 3 docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.now):

image

The changes in this branch are consistent with "the recommended way" shown in the red box on that page.

eecavanna:
I confirmed that, even in the Python **3.10** docs, this replacement is "recommended."

image

The reason I checked that version of the Python docs is that that is the Python version that is running in the container image that runs the FastAPI app, according to [its Dockerfile](https://github.com/microbiomedata/nmdc-runtime/blob/2eb01b0b7a27b3fbfa148d2e6f72da5727a17b47/nmdc_runtime/fastapi.Dockerfile#L2C6-L2C31).

eecavanna:
I'm comfortable with this branch being merged into `main` as is.

dwinston (approved):
Great. Thanks.

Issue #797, opened 2024-11-26
API endpoint `"/nmdcschema/{collection_name}"` takes 4x longer when using pagination (with `functional_annotation_agg`)


eecavanna:
The following HTTP request currently takes 1.2 minutes to complete in the dev environment:

```sh
curl -X 'GET' \
  'https://api-dev.microbiomedata.org/nmdcschema/functional_annotation_agg?filter=%7B%22was_generated_by%22%3A%7B%22%24regex%22%3A%22wfmp%22%7D%7D&max_page_size=50&projection=was_generated_by' \
  -H 'accept: application/json'
```
> Notice: The `max_page_size` is `50` (i.e. not `0`).

Here's a screenshot showing the same parameter values entered into Swagger UI:

<img width="392" alt="image" src="https://github.com/user-attachments/assets/ae5c4876-da21-48b2-93f3-3f6f2d53b2ef">

In contrast, the following HTTP request currently takes 0.3 minutes to complete in the dev environment:

```sh
curl -X 'GET' \
  'https://api-dev.microbiomedata.org/nmdcschema/functional_annotation_agg?filter=%7B%22was_generated_by%22%3A%7B%22%24regex%22%3A%22wfmp%22%7D%7D&max_page_size=0&projection=was_generated_by' \
  -H 'accept: application/json'
```
> Notice: The `max_page_size` is `0`.

I suspect the issue is related to the fact that the function that powers that endpoint does different things depending upon whether `max_page_size` is `0` or not `0`, as shown below. I notice, in the case of the former, the Mongo querying code does not use the `sort`, `limit`, or `allow_disk_use` kwargs.

https://github.com/microbiomedata/nmdc-runtime/blob/d7da63fdfaf9262ad1da4d0d9c4301b8d119b534/nmdc_runtime/api/endpoints/util.py#L104-L134

Thank you, @kheal, for reporting this issue. In the case of what she was working on today, setting `max_page_size=0` was a sufficient workaround. In other cases, it may not be (e.g. when the response would contain more data than some piece of our infrastructure—or the requesting HTTP client—supports).

eecavanna:
Reopening. Although the `was_generated_by` index now exists in the development database, the original symptom persists.

eecavanna:
Here are the times I measured today, with the `was_generated_by` index in place.

This request/response cycle, with `max_page_size=50`, took 1 minute and 34 seconds to complete:

```shell
date && curl -X 'GET' \
[...]

PR #803, opened 2024-11-30
Update `/nmdcschema/{collection_name}` API endpoint description to include performance-related guidance

eecavanna:
In this branch, I updated documentation about an API endpoint and a helper function that it calls.

### Details

I updated user-facing documentation about an API endpoint in an attempt to help users avoid performance-related surprises and to more easily work around them if they do encounter them. This documentation gets displayed on Swagger UI.

I also updated some internal documentation (a docstring) related to a function called by that endpoint.

Finally, I added some `TODO` comments related to telling the user what all the valid collection names are. We currently send them to the schema docs for that. We could, instead, implement an API endpoint that returns a list. That would allow us to refrain from giving the nuanced answer of "It's the names of the slots of the `Database` class, as long as those slots are multivalued and inlined_as_list." The API endpoint we implement could check the latter conditions instead of making that the user's responsibility.

Props to @dwinston for bringing the MongoDB concept of a "prefix expression" to my attention.

### Related issue(s)

Related to #797 

The reported performance issue still persists. This PR branch does not include a fix—just some user-facing guidance and internal documentation related to it.

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I spin up my local development environment, viewed the endpoint's updated documentation on Swagger UI (shown here), and confirmed it looked the way I expected.

image

For comparison, here's what the documentation used to be:

image

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/nmdc_runtime/api/endpoints/nmdcschema.py
+++ b/nmdc_runtime/api/endpoints/nmdcschema.py
@@ -116,10 +116,18 @@ def list_from_collection(
     mdb: MongoDatabase = Depends(get_mongo_db),
 ):
     """
-    The GET /nmdcschema/{collection_name} endpoint is a general purpose way to retrieve metadata about a specified
-    collection given user-provided filter and projection criteria. Please see the [Collection Names](https://microbiomedata.github.io/nmdc-schema/Database/)
-    that may be retrieved. Please note that metadata may only be retrieved about one collection at a time.
+    Returns resources that match the specified filter criteria, residing in the specified database collection.
+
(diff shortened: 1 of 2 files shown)


eecavanna:
Merging this documentation-only PR branch without review (beyond a thumbs up on a [comment](https://github.com/microbiomedata/nmdc-runtime/issues/797#issuecomment-2508530750) where I said I intended to make these changes) in order to simplify the merging of [a more substantial PR branch](https://github.com/microbiomedata/nmdc-runtime/pull/804) soon, since both PR branches contain changes to the same file.

PR #812, opened 2024-12-05
Display copy-paste HTTP Authorization header with Bearer token on Swagger UI for ORCiD-Authenticated user

### LINKED ISSUE #742: Expose Auth Token to API user for CLI Client
From @shreddd in [Slack](https://nmdc-group.slack.com/archives/C06GJ52E4T1/p1729545493384969)

> We need to address a programmatic access use case from JGI and I wanted to understand how best to support this. The simple (hopefully) thing we would like short term is being able to quickly get the access token (bearer token) that an API user can use in a CLI client. Right now it requires a somewhat hacky approach to introspect the outgoing call or looking at the Curl dump of the call. 

Related to #404

dwinston:
closes #742

In this branch, I used the Swagger UI's `onComplete` parameter to display bearer token info if present as a cookie (e.g. via ORCiD login).

### Details

image

### Related issue(s)

Closes #742
...

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by manual inspection of `http://127.0.0.1:8000/docs`

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/nmdc_runtime/api/main.py
+++ b/nmdc_runtime/api/main.py
@@ -488,10 +488,7 @@ def custom_swagger_ui_html(
         onComplete += f"""
             ui.preauthorizeApiKey('bearerAuth', '{access_token}');
             token_info = document.createElement('section');
-            token_info.classList.add('nmdc-info');
-            token_info.classList.add('nmdc-info-token');
-            token_info.classList.add('block');
-            token_info.classList.add('col-12');
+            token_info.classList.add('nmdc-info', 'nmdc-info-token', 'block', 'col-12');
             token_info.innerHTML = <double-quote>
                 <p>You are now authorized. Prefer a command-line interface (CLI)? Use this header for HTTP requests:</p>
                 <p>
@@ -514,10 +511,7 @@ def custom_swagger_ui_html(
         info_banner_innerhtml = os.getenv("INFO_BANNER_INNERHTML")
         onComplete += f"""
             banner = document.createElement('section');
-            banner.classList.add('nmdc-info');
-            banner.classList.add('nmdc-info-banner');
-            banner.classList.add('block');
-            banner.classList.add('col-12');
+            banner.classList.add('nmdc-info', 'nmdc-info-banner', 'block', 'col-12');
             banner.innerHTML = `{info_banner_innerhtml.replace('"', '<double-quote>')}`;
             document.querySelector('.information-container').prepend(banner);
         """.replace(
--- a/nmdc_runtime/api/main.py
+++ b/nmdc_runtime/api/main.py
@@ -510,7 +510,9 @@ def custom_swagger_ui_html(
                 </p>
             </double-quote>;
             document.querySelector(<double-quote>.information-container</double-quote>).append(token_info);
-        """.replace("\n", " ")
+        """.replace(
+            "\n", " "
+        )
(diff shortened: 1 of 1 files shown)


shreddd:
I think this is a reasonable solution. In the interest of security best practice can add a display toggle that will turn this on/off on the user's screen?

eecavanna:
Hi @dwinston, @shreddd and I looked at this branch together during today's hackathon (and we both got it up and running locally). I will take a shot at implementing the `***` masking that @shreddd described. I'll post another message here when I'm done (whether I succeeded or failed—the potential hurdle being understanding how the JS-in-PY gets transformed between the Python context and the final web page, and at what point the HTML-in-JS elements eventually get mounted to the DOM).

shreddd (changes requested):
Add a show hide/toggle similar to https://data.microbiomedata.org/user
Screen Shot 2024-12-06 at 2 14 10 PM

eecavanna:
Hi @shreddd, here's what I have added:

The token is masked by default:

image

When the user clicks "Show token", it gets unmasked (I've blurred it in my screenshot tool):

image

When the user clicks "Copy token," the token (intentionally not the entire header) gets copied to their clipboard and a success message appears. If their browser doesn't support copying, an error message will appear instead.

image

What do you think?

nmdc_runtime/api/main.py
```diff
@@ -510,9 +545,67 @@ def custom_swagger_ui_html(
...
+                    .nmdc-info {{\
+                        padding: 1em;\
+                        background-color: #448aff1a;\
+                        border: .075rem solid #448aff;\
+                    }}\
+                    .nmdc-info-token code {{\
+                        font-size: x-small;\
+                    }}\
+                </style>\
+            </head>""",
```
eecavanna, line 561: I didn't realize this snippet was added as part of this same PR. I will remove the unnecessary trailing backslashes now.

eecavanna:
OK, I'm done working on this PR (I have closed `nmdc-runtime` in my IDE, shut down my Docker Compose stack, and am going to context switch to another task).

eecavanna (approved):
I validated the functionality locally.

If I hadn't just posted a message saying I was going to switch context, I think I would take a few minutes to convert `<double-quote>` into literal single quote characters, and also revisit whether those trailing backslashes are necessary in all the places they currently exist (given that they were not necessary in the second argument to `.replace()`). If this gnaws at me, I'll come back and do that stuff (I think it'll facilitate future maintenance). Either way, I'm comfortable with this being merge din as is.

eecavanna:
OK, I'm done addressing what I wanted to on this branch. I'm comfortable with this being merges in as is, assuming @shreddd is comfortable with the masking/copying behavior. I'm going to shift my focus away from this PR now.

eecavanna (approved):

eecavanna:
I'll merge this in now. It:
- Displays the token
- Masks the token by default
- Allows the user to unmask the token (by clicking a button)
- Allows the user to copy the token to their clipboard (by clicking a button)
- Shows a message saying whether the copy operation was successful

Risks:
- Stakeholders are not satisfied with the masking behavior

I am comfortable with that risk, given the narrow scope of these changes and the fact that there's still a week until `main` will be released to production (in case any stakeholders want to roll this back).

PR #842, opened 2024-12-13
Remove obsolete guidance from development guide

shreddd:
`export DOCKER_DEFAULT_PLATFORM=linux/amd64` was an older recommendation to allow amd64 containers to run on M1/M2 macs. However newer versions of Docker Desktop don't like this, and you should just let Docker pull the right container,


eecavanna (approved):
LGTM - thanks for updating the documentation!

PR #856, opened 2024-12-26
Add class with helper methods to fill in missing records in database

### LINKED ISSUE #756: Modify GOLD ETL code to check existing records and only create missing records.
**Is your feature request related to a problem? Please describe.**
We have studies that will be in a hybrid state where we have biosamles from submission portal to mongo ETL and we need to make corresponding data_generation_set records. 

**Describe the solution you'd like**
ETL code should check existing records for a study, determine what records are missing and make those. Example project is 
https://github.com/microbiomedata/issues/issues/813

**Acceptance Criteria**
Updated GOLD ETL code that can be run to makes data_generation_set records for https://github.com/microbiomedata/issues/issues/813

sujaypatil96:
We are trying to address the general problem of "database stitching" in this PR. It will often be the case that we have certain records in the database (say biosamples), but records that are needed for the next bits of processing of these "entities" might be missing.

For example, the first and most common use case that we are trying to address is filling in missing data_generation_set records for biosamples that have been brought into the database through the `SubmissionPortalTranslator` mechanism.

- [x] Add Dagster harness to support calling of helper methods in `DatabaseUpdater` class (general interface which we will use to collect helper methods to facilitate all use cases of "database stitching")
- [x] Add helper methods to class
  - [x] Create missing DataGeneration records given Biosamples that are already present in the database

### Details

...

### Related issue(s)

Fixes #756 https://github.com/microbiomedata/issues/issues/951 https://github.com/microbiomedata/issues/issues/813

### Related subsystem(s)

- [ ] Minter
- [x] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by...

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/nmdc_runtime/site/graphs.py
+++ b/nmdc_runtime/site/graphs.py
@@ -57,6 +57,9 @@ from nmdc_runtime.site.ops import (
     get_ncbi_export_pipeline_inputs,
     ncbi_submission_xml_from_nmdc_study,
     ncbi_submission_xml_asset,
+    get_database_updater_inputs,
+    nmdc_study_id_filename,
+    missing_data_generation_repair,
 )
 from nmdc_runtime.site.export.study_metadata import get_biosamples_by_study_id
 
@@ -467,3 +470,14 @@ def nmdc_study_to_ncbi_submission_export():
         all_instruments,
     )
     ncbi_submission_xml_asset(xml_data)
+
+
+@graph
+def fill_missing_data_generation_data_object_records():
+    study_id = get_database_updater_inputs()
+    database = missing_data_generation_repair(study_id)
+
+    database_dict = nmdc_schema_object_to_dict(database)
+    filename = nmdc_study_id_filename(study_id)
+    outputs = export_json_to_drs(database_dict, filename)
+    add_output_run_event(outputs)
--- a/nmdc_runtime/site/graphs.py
+++ b/nmdc_runtime/site/graphs.py
@@ -474,8 +474,10 @@ def nmdc_study_to_ncbi_submission_export():
 
 @graph
 def fill_missing_data_generation_data_object_records():
-    study_id = get_database_updater_inputs()
-    database = missing_data_generation_repair(study_id)
+    (study_id, gold_nmdc_instrument_mapping_file_url) = get_database_updater_inputs()
+    gold_nmdc_instrument_map_df = get_df_from_url(gold_nmdc_instrument_mapping_file_url)
+
+    database = missing_data_generation_repair(study_id, gold_nmdc_instrument_map_df)
 
     database_dict = nmdc_schema_object_to_dict(database)
     filename = nmdc_study_id_filename(study_id)
--- a/nmdc_runtime/site/ops.py
+++ b/nmdc_runtime/site/ops.py
@@ -1255,19 +1255,49 @@ def nmdc_study_id_filename(nmdc_study_id: str) -> str:
 
 
 @op(
-    config_schema={"nmdc_study_id": str},
-    out={"nmdc_study_id": Out(str)},
+    config_schema={
+        "nmdc_study_id": str,
+        "gold_nmdc_instrument_mapping_file_url": str,
+    },
+    out={
+        "nmdc_study_id": Out(str),
+        "gold_nmdc_instrument_mapping_file_url": Out(str),
+    },
 )
-def get_database_updater_inputs(context: OpExecutionContext) -> str:
-    return context.op_config["nmdc_study_id"]
+def get_database_updater_inputs(context: OpExecutionContext) -> Tuple[str, str]:
+    return (
+        context.op_config["nmdc_study_id"],
+        context.op_config["gold_nmdc_instrument_mapping_file_url"],
+    )
 
 
-@op(required_resource_keys={"runtime_api_user_client"})
+@op(
+    required_resource_keys={
+        "runtime_api_user_client",
+        "runtime_api_site_client",
+        "gold_api_client",
+    }
+)
 def missing_data_generation_repair(
-    context: OpExecutionContext, nmdc_study_id: str
+    context: OpExecutionContext,
+    nmdc_study_id: str,
+    gold_nmdc_instrument_map_df: pd.DataFrame,
 ) -> nmdc.Database:
-    client: RuntimeApiUserClient = context.resources.runtime_api_user_client
-    database_updater = DatabaseUpdater(client, nmdc_study_id)
-    database = database_updater.get_database()
+    runtime_api_user_client: RuntimeApiUserClient = (
+        context.resources.runtime_api_user_client
+    )
+    runtime_api_site_client: RuntimeApiSiteClient = (
+        context.resources.runtime_api_site_client
+    )
+    gold_api_client: GoldApiClient = context.resources.gold_api_client
+
+    database_updater = DatabaseUpdater(
+        runtime_api_user_client,
+        runtime_api_site_client,
+        gold_api_client,
+        nmdc_study_id,
+        gold_nmdc_instrument_map_df,
+    )
+    database = database_updater.create_missing_dg_records()
 
     return database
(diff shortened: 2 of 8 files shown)


nmdc_runtime/site/repair/database_updater.py
```diff
@@ -0,0 +1,151 @@
...
+    def _fetch_gold_projects(self, gold_biosample_id: str):
+        """Fetch response from GOLD /projects API for a given biosample id.
+
+        :param gold_biosample_id: GOLD biosample ID
+        :return: Dictionary containing the response from the GOLD /projects API.
+        """
+        return self.gold_api_client.fetch_projects_by_biosample(gold_biosample_id)
+
+    @lru_cache
+    def create_missing_dg_records(self):
```
aclum, line 62: this function is missing logic to handle a biosample records having multiple GOLD biosample identifiers.

nmdc_runtime/site/resources.py
```diff
@@ -134,7 +134,7 @@ def get_biosamples_for_study(self, study_id: str):
             f"/queries:run",
```
aclum, line 134: recommend rewritting this as a query against nmdcschema/{collectionname} endpoint (https://api.microbiomedata.org/docs#/metadata/list_from_collection_nmdcschema__collection_name__get). queries:run does not support setting page size or pagination so may produce truncated results. See [#434 queries:run endpoint truncating results](https://github.com/microbiomedata/nmdc-runtime/issues/434)

nmdc_runtime/site/repair/database_updater.py
sujaypatil96: This concern has been addressed in 128906f46f21f1a21eed4841123b078b99c47177. See https://github.com/microbiomedata/nmdc-runtime/compare/main...issue-756#diff-3476d6b3affc1cc68b5f2a6fec08132f1c4d7add65e55b116b6baf9baa57a262R82-R91

nmdc_runtime/site/resources.py
sujaypatil96: Addressed in https://github.com/microbiomedata/nmdc-runtime/compare/main...issue-756#diff-ad7656740b805d4bb169fca6949ae16074bd39d7db7f22c16e2ed7229bdfaf3bR131-R148

I've also added comments about needing to improve that method in the future. We are currently assigning one of the parameters of that endpoint, `max_page_size` an arbitrarily high number [10000] so that we don't have to make multiple requests to walk through the paginated results returned by that endpoint when the `max_page_size` is used with default parameter values.

aclum (dismissed):

nmdc_runtime/site/graphs.py
```diff
@@ -467,3 +470,16 @@ def nmdc_study_to_ncbi_submission_export():
         all_instruments,
     )
     ncbi_submission_xml_asset(xml_data)
+
+
+@graph
+def fill_missing_data_generation_data_object_records():
```
pkalita-lbl, line 476: To me "fill" implies that it's going to actually commit the results to Mongo. Is there any reason this Graph doesn't do that?

nmdc_runtime/site/ops.py
```diff
@@ -1241,3 +1246,58 @@ def ncbi_submission_xml_from_nmdc_study(
         all_instruments,
     )
     return ncbi_xml
+
+
+@op
+def nmdc_study_id_filename(nmdc_study_id: str) -> str:
```
pkalita-lbl, line 1252: I think this needs a more specific name since it produces a very specific file name.

nmdc_runtime/site/repair/database_updater.py
```diff
@@ -0,0 +1,150 @@
...
+from nmdc_runtime.site.resources import (
+    RuntimeApiUserClient,
+    RuntimeApiSiteClient,
+    GoldApiClient,
+)
+from nmdc_runtime.site.translation.gold_translator import GoldStudyTranslator
+from nmdc_schema import nmdc
+
+
+class DatabaseUpdater:
```
pkalita-lbl, line 13: Is there any value in this being a class? Since it's not inheriting from anything and the internal state never really gets modified, could `create_missing_dg_records` just be a top-level function instead?

nmdc_runtime/site/repair/database_updater.py
```diff
@@ -0,0 +1,150 @@
...
+    @lru_cache
+    def _fetch_gold_projects(self, gold_biosample_id: str):
+        """Fetch response from GOLD /projects API for a given biosample id.
+
+        :param gold_biosample_id: GOLD biosample ID
+        :return: Dictionary containing the response from the GOLD /projects API.
+        """
+        return self.gold_api_client.fetch_projects_by_biosample(gold_biosample_id)
+
+    def create_missing_dg_records(self):
```
pkalita-lbl, line 61: Similar to the above comment, I'm a little hesitant about the "missing" part of this name. To me this is `generate_data_generation_set_from_gold_api`.

nmdc_runtime/site/repository.py
```diff
@@ -922,6 +923,54 @@ def biosample_export():
...
+                                "password": {"env": "GOLD_API_PASSWORD"},
+                            },
+                        },
+                    },
+                ),
+                "ops": {
+                    "get_database_updater_inputs": {
+                        "config": {
+                            "nmdc_study_id": "",
+                            "gold_nmdc_instrument_mapping_file_url": "https://raw.githubusercontent.com/microbiomedata/nmdc-schema/refs/heads/main/assets/misc/gold_seqMethod_to_nmdc_instrument_set.tsv",
```
pkalita-lbl, line 964: You might have already explained this to me and I've forgotten, but why do we keep this file in the `nmdc-schema` repository?

aclum (approved):

pkalita-lbl (approved):
Overall this looks nice and I appreciate the documentation you've added. I have some comments and questions, but none of it is so critical that it needs to hold up merging.

---

I guess my main concern, design-wise, is that I'd prefer to be more explicit in terms of naming things and avoid vague (and somewhat misleading) wording like "missing" or "repair". 

For example, there's a Dagster Op named `missing_data_generation_repair`. What it actually seems to do is produce `NucleotideSequencing` records based on information fetched from the GOLD API for a given `Study`. That's all well and good. But it would do the same thing whether or not there are existing `NucleotideSequencing` records associated with the given `Study`. So the "missing" part of the name seems misleading. Similarly, to me, the word "repair" implies that the `Study` input is going to be modified in some way. It also implies that the study was "broken" before 😂. All that being said, I probably would have named it something more explicit like `generate_data_generation_set_from_gold_api_for_study`. It's long, but, hey, characters are free.

nmdc_runtime/site/repair/database_updater.py
sujaypatil96: Makes sense! "missing" is a little misleading, "generate" is better.

nmdc_runtime/site/repair/database_updater.py
sujaypatil96: Hm, there will be other methods that I add to this "interface" (class) in the future, so we have to decide between "collecting" all of these methods under that class or have them be top level functions?

nmdc_runtime/site/ops.py
sujaypatil96: Will address it!

nmdc_runtime/site/graphs.py
sujaypatil96: Hm, so we want to see the JSON export that this job produces, visually QC it and then ingest it into the system. So it's not doing the "filling" automatically. Is "stitch_" a good prefix?

sujaypatil96:
Awesome!! this is such a wonderful review!! thank you @pkalita-lbl. I have another PR coming out next week that adds more functionality to this module, I'll make sure to address all the review comments you've left here in a commit on that PR if that works?

PR #861, opened 2025-01-16
Fix bug where `validate_json` referential integrity check ignored documents in database

### LINKED ISSUE #860: Real-time referential integrity checking is broken
Real-time referential integrity checking was introduced into the `validate_json` function via PR https://github.com/microbiomedata/nmdc-runtime/pull/835 this week.

The implementation is broken.

In terms of behavior visible to end users:
- It does not take into account data that already exists in the database (❗). As a result, it reports false positives.

In terms of the code:
- The `get_collection` method that was added to the `OverlayDB` class only returns the so-called "top" database (it does not return a hybrid top+bottom database).
- An `OverlayDB` instance is being passed to a function that was designed to receive an instance of pymongo's `Database` class.

CC: @aclum , @shreddd , @dwinston

eecavanna:
On this branch, I fixed a bug where the real-time referential integrity checker in the `validate_json` function was ignoring documents that were already in the database.

### Details

On this branch, I changed the real-time referential integrity checker code to no longer use the `OverlayDB` class. When I first wrote the real-time referential integrity checker code, I had made an incorrect assumption about how the `OverlayDB` class worked.

Instead of it using the `OverlayDB` class, I updated the checker to use the real database directly and to use the JSON payload directly. Checking now happens in two stages (one for each of those potential "residences" of each referenced document).

The two stages are (in short—as a more detailed explanation is in the code comments):
1. Check the database
2. If the document isn't found there, check the JSON payload

With those changes in place, the checker no longer ignores documents that already exist in the database. I added a unit test that demonstrates this.

### Related issue(s)

Fixes #860

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by adding some unit tests that target them, and ensuring that those tests pass. I will leave it to GitHub Actions to tell me whether all pre-existing tests still pass.

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/nmdc_runtime/api/endpoints/metadata.py
+++ b/nmdc_runtime/api/endpoints/metadata.py
@@ -168,10 +168,13 @@ async def get_stored_metadata_object(
 
 @router.post("/metadata/json:validate", name="Validate JSON")
 async def validate_json_nmdcdb(docs: dict, mdb: MongoDatabase = Depends(get_mongo_db)):
-    """
-
+    r"""
     Validate a NMDC JSON Schema "nmdc:Database" object.
 
+    This API endpoint validates the JSON payload in two steps. The first step is to check the format of each document
+    (e.g., the presence, name, and value of each field). If it encounters any violations during that step, it will not
+    proceed to the second step. The second step is to check whether all documents referenced by the document exist,
+    whether in the database or the same JSON payload. We call the second step a "referential integrity check."
     """
 
     return validate_json(docs, mdb, check_inter_document_references=True)
--- a/nmdc_runtime/util.py
+++ b/nmdc_runtime/util.py
@@ -542,6 +542,13 @@ class OverlayDB(AbstractContextManager):
     overlay collection, that id is marked as "seen" and will not also be returned when
     subsequently scanning the (unmodified) base-database collection.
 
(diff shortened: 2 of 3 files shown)


eecavanna:
Props to @mbthornton-lbl for reporting this issue via [Slack](https://nmdc-group.slack.com/archives/C077QG7C1V2/p1736979878086569) earlier today—thank you!

eecavanna:
I'm ready for this to be merged in, even in my absence (in case the reviews take place before I'm online).

aclum (approved):

shreddd (approved):
This looks good. Possible future enhancements:
- We should also look at optimizations to queries / caching / memoization as a way to speed up multiple queries looking for the same doc. 
- Seeing if we can do this as a transaction that can be rolled back.

eecavanna:
Thanks, @shreddd!

Regarding:
> Seeing if we can do this as a transaction that can be rolled back.

I want to expand on that by saying (as we discussed over Zoom earlier today) the transaction would allow us to "stage" the proposed operations on the real database (i.e. "perform" them, but without committing the transaction), do the referential integrity scan on that single database (on which the operations have been "staged"), and then abort the transaction.

I'll merge this in now in order to fix the bug currently present in the development environment.

PR #910, opened 2025-02-15
Upgrade MongoDB to v8 and consolidate Dockerfiles into multi-stage Dockerfile

### LINKED ISSUE #880: Update docker compose configs to use mongo:7.0.15 image rather than mongo:6.0.4
- [ ] Shadows https://github.com/microbiomedata/infra-admin/issues/160
- [ ] Do after successful production deploy with https://github.com/microbiomedata/issues/issues/988

### LINKED ISSUE #909: Consolidate recurring Dockerfile portions into shared base image
The top portions of all of the Dockerfiles in the repo are mostly identical to one another. When rebuilding the images simultaneously via Docker Compose, the same steps (e.g. downloading an installing Python dependencies) are happening 3 times. The task here is to consolidate the Dockerfile steps into a single base image so those steps happen only once.

eecavanna:
On this branch, I made it so **building** the Docker Compose stack is faster. I also **updated MongoDB** to version 8.

#### Speeding up builds

I consolidated the three Dockerfiles (`dagster.Dockerfile`, `fastapi.Dockerfile`, and `test.Dockerfile`) into one, multi-stage `Dockerfile`. I then updated the `docker-compose.yml` and `docker-compose.test.yml` files accordingly.

The speed boost came from the fact that running `docker compose up --build` (and similar commands) cause all images to be built in parallel. Although the three Dockerfiles started with a similar sequence of steps, since they were being built in parallel, they were not able to take advantage of cached layers being generated by each other.

So, building them involved downloading and installing system-level dependencies three times simultaneously, downloading and installing Python `main.txt` dependencies three times simultaneously, etc. Now, those things only happen once, and the three late-stage images _reuse the result_ (instead of creating their own copies in parallel). Each late-stage image _does_ still have a few custom steps, but they all share a common base.

###### [Before](https://github.com/microbiomedata/nmdc-runtime/actions/runs/13446316336/job/37572355642) (randomly chosen)

Sample one: 4m 51s

image

Sample two: 5m 10s

image

###### [After](https://github.com/microbiomedata/nmdc-runtime/actions/runs/13558695330/job/37897826788) (on this branch; note that I divided one of the workflow steps into three to help with debugging)

Sample one: 3m 38s

image

Sample two: 4m 3s

image

#### Updating Mongo

I updated the MongoDB version in the local development/test stacks to match that used in the production environment (in terms of version number, not in terms of the container image variant; i.e. Bitnami-maintained versus community-maintained).

Note: The MongoDB servers in the local development/test environment are not running in replica set mode—there will be a future PR about that.

### Related issue(s)

Fixes #909 
Fixes #880
Fixes https://github.com/microbiomedata/infra-admin/issues/160

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [x] Other

### Testing

I tested them by (a) confirming all tests pass locally, (b) confirming all tests pass on GHA, (c) confirming I can spin up the local development stack and access the Swagger UI and Dagster UI.

### Documentation

- [ ] Other (explain below)

### Maintainability


eecavanna:
**Heads up** to recent Runtime committers (e.g. @dwinston, @shreddd, @sujaypatil96, @pkalita-lbl):

On the latest commit on this branch; when I spun up my local development stack (via `$ make up-dev`), my `mongo` container did not start up completely (I think it went into a reboot loop). It was only after I deleted it, deleted its volume (which was named `nmdc-runtime_nmdc_runtime_mongo_data`), and then recreated it (via `$ docker compose up --detach mongo`), that it started up completely.

I suspect the issue was due to my having upgraded its container image from version 6 directly to version 8 (instead of from 6 to 7 to 8, as is recommended in the Mongo documentation).

eecavanna:
On this branch, I forgot to update the GHA workflow that builds the container images and pushes them to the container image registries.  I have a fix on my laptop, but my laptop's Internet connection is down. I will push up the fix once my laptop is back online. 

The fix involved (both in the workflow's YAML file) removing a variable from a reference to a Dockerfile, and adding the `target` option to a step.

eecavanna:
Based on the two pairs of workflow runs I compared, I believe the changes on this branch speed up the build by about a minute (i.e. by about 20%).

I'm ready for this PR to be reviewed/merged.

shreddd (approved):
Consolidating to a multi-stage build makes a lot of sense

PR #998, opened 2025-05-23
Update `/queries:run` API endpoint so only users having specific allowance can run aggregation command

### LINKED ISSUE #970: Update `/queries:run` API endpoint so only users having specific allowance can run `aggregation` command
The idea of defining a `_runtime.api.allow` action specific to the "aggregation" command on the `/queries:run` API endpoint came up during a recent conversation about the fact that "aggregations" can be computationally expensive and, currently, anyone that has an ORCID account can perform one.

The task here is to:
- [x] 1. **Decide** whether we will introduce this new "action" (non-technical work); and, if so...
- [x] 2. **Introduce** it, using an existing "action" as a reference (technical work)
- [x] 3. **Grant** it to all existing Runtime accounts so there is no immediate change from their perspective (technical work)
- [ ] 4. **Decide** whether we will revoke it from certain existing Runtime accounts (non-technical work)

In practice, someone might introduce it in a PR before the decision has been made, in an attempt to streamline the decision process (by having an example to look at).

hesspnnl:
On this branch, I edited /queries:run to check if a user trying to aggregate has the correct permissions. 

...

### Related issue(s)

Fixes #970 

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by running them locally and confirming they all pass. 

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/tests/test_api/test_endpoints.py
+++ b/tests/test_api/test_endpoints.py
@@ -1559,7 +1559,13 @@ def test_run_query_aggregate__first_batch_and_its_cursor_id(api_user_client):
     faker = Faker()
     studies = faker.generate_studies(6, title=study_title)
     study_set.insert_many(studies)
-
+    mdb = get_mongo_db()
+    # give user permission to run aggregate queries
+    allow_spec = {
+        "username": api_user_client.username,
+        "action": "/queries:run(query_cmd:AggregateCommand)",
+    }
+    mdb["_runtime.api.allow"].replace_one(allow_spec, allow_spec, upsert=True)
     # Test case 1: When the first batch is empty, its `cursor.id` is null.
     response = api_user_client.request(
         "POST",
@@ -1676,7 +1682,12 @@ def test_run_query_aggregate__second_batch_and_its_cursor_id(api_user_client):
     faker = Faker()
     studies = faker.generate_studies(6, title=study_title)
     study_set.insert_many(studies)
-
+    # give user permission to run aggregate queries
+    allow_spec = {
+        "username": api_user_client.username,
+        "action": "/queries:run(query_cmd:AggregateCommand)",
+    }
+    mdb["_runtime.api.allow"].replace_one(allow_spec, allow_spec, upsert=True)
     # Test case 1: When the second batch is empty, its `cursor.id` is null.
     response = api_user_client.request(
         "POST",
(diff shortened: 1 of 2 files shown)


tests/test_api/test_endpoints.py
```diff
@@ -987,6 +987,74 @@ def test_run_query_update_as_user(api_user_client):
             },
         )
 
+def test_run_query_aggregate_as_user(api_user_client):
+    """
+    Submit a request to aggregate data without the correct permissions. Then add the permissions
+    and submit the same request again, this time expecting a successful response.
```
eecavanna, line 993: I like this game plan for the test!

tests/test_api/test_endpoints.py
```diff
@@ -987,6 +987,74 @@ def test_run_query_update_as_user(api_user_client):
...
+    # and insert them into the database.
+    #
+    # Note: The reason assertion is necessary is that some longstanding tests
+    #       in this repostory leave "residue" in the test database after they
+    #       run. See "FIXME" comments in this module for more details.
+    #
+    study_title = "My study"
+    nonexistent_study_title = "Nonexistent study"
+    assert study_set.count_documents({"title": study_title}) == 0
+    assert study_set.count_documents({"title": nonexistent_study_title}) == 0
```
eecavanna, line 1009: I think the three regular studies will be sufficient for this test, since the test is focused on the endpoint's permissions as opposed to how the endpoint handles existing versus nonexistent documents.

```suggestion
    study_title = "My study"
    assert study_set.count_documents({"title": study_title}) == 0
```

tests/test_api/test_endpoints.py
```diff
@@ -987,6 +987,74 @@ def test_run_query_update_as_user(api_user_client):
...
+                    "$match": {
+                        "title": nonexistent_study_title,
+                    },
+                },
+            ],
+            "cursor": {"batchSize": 5},
+        },
+    )
+    assert response.status_code == 200
+
```
eecavanna, line 1057: Looking great! One thing missing from this test is a "cleanup" step, where we delete the documents we created during the test. In other projects, this may be handled automatically by test fixtures. In this project, it has not been that way until just a few days ago (that's why lots of tests currently have their own "cleanup" code at the end. I'll walk you through the fixture approach next time we pair program.

tests/test_api/test_endpoints.py
```diff
@@ -1559,7 +1627,13 @@ def test_run_query_aggregate__first_batch_and_its_cursor_id(api_user_client):
...
     studies = faker.generate_studies(6, title=study_title)
     study_set.insert_many(studies)
-
+    mdb = get_mongo_db()
+    # give user permission to run aggregate queries
+    allow_spec = {
+        "username": api_user_client.username,
+        "action": "/queries:run(query_cmd:AggregateCommand)",
+    }
+    mdb["_runtime.api.allow"].replace_one(allow_spec, allow_spec, upsert=True)
```
eecavanna, line 1636: We can extract this recurring blurb into a helper function (or into another fixture). I'll elaborate on this next time we pair program.

eecavanna (approved):
LGTM!

eecavanna:
Oops! I forgot to grant our existing users this new permission. I will do that later today, in our Spin-hosted development environment.

PR #999, opened 2025-05-23
Customize host port mapping via `.env` file

### LINKED ISSUE #987: Run both `-dev` and `-test` `docker compose` container stacks on the same machine
> I would like to keep my dev environment up, because its happy and green and I need it to show status reports, but I see that the CI/CD tests are failing in my PR (in the test environment, I believe) and I want to in parallel debug why locally.

**Goal solution**

Run both `make up-dev` and `make up-test` at the same time on the same machine.

```
AS A user,
INSTEAD OF not being able to run both stacks at the same time,
I WANT TO run both
SO THAT I can develop and debug in parallel locally.
```

**Acceptance criteria**

Can run both `make up-dev` and `make up-test` at the same time on the same machine.

**Alternatives considered**

Manually change port assignments locally, playing whack-a-~~mole~~port if/when encountering port assignment conflicts.

**Additional context**

[Thread in `#infra-admin` channel on the NMDC team Slack workspace](https://nmdc-group.slack.com/archives/C05EFGHCYRH/p1746025407882439).

/cc @sierra-moxon @pkalita-lbl

dwinston:
On this branch, I added to `.env.example` and used variable substitution in `docker-compose*.yml` files.

### Related issue(s)

Closes #987

### Related subsystem(s)

- [x] Other

local dev setup.

### Testing

I tested these changes by `make test`.

### Documentation

- [x] Other (explain below)

No doc updates strictly needed, as `dev`-stack defaults remain as is.

### Maintainability

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -57,3 +57,13 @@ MONGO_REMOTE_DUMP_DIR=/global/cfs/projectdirs/m3408/nmdc-mongodumps/dump_nmdc-pr
 # Note: This is currently only used by the `nersc-sshproxy` target in `Makefile`.
 # Reference: https://docs.nersc.gov/connect/mfa/#sshproxy
 PATH_TO_NERSC_SSHPROXY=/Users/Alice/Downloads/sshproxy
+
+# `make up-dev` and `make up-test` will map services to host-machine port numbers using the following environment
+# variables in your `.env` file. This way, different developers can use different local port numbers, without modifying
+# files that get stored in the repository.
+DEV_STACK_HOST_MACHINE_PORT_DAGIT=3000
+DEV_STACK_HOST_MACHINE_PORT_FASTAPI=8000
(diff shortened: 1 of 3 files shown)


eecavanna (approved):
Looks good to me. Thanks for implementing this.

I have two comments. I am not posting them as formal "GitHub diff suggestions" because I don't want to give them any extra "weight".

1. I already think of the port mappings as "host port: container port" (as opposed to "host machine port:container port")—although "host machine" (and "local machine") is a phrase I use in life—and so would be OK with the environment variable names not including the "`_MACHINE`" substring.
2. The references to environment variables (in the Docker Compose file) can include [default values](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#interpolation-syntax), like this: `- "${TEST_STACK_HOST_MACHINE_PORT_DAGIT:-3000}:3000"`

sierra-moxon (approved):
Thank you Donny!

PR #1001, opened 2025-05-28
Add referential integrity check to "delete" command of `POST /queries:run` API endpoint

### LINKED ISSUE #1000: Add referential integrity check to `POST /queries:run` API endpoint (for `delete` command)

eecavanna:
On this branch, I added a real-time referential integrity check to the `POST /queries:run` API endpoint; specifically, to a branch of code that handles the "delete" command.

Before, the endpoint allowed users to delete documents that other documents referred to, even when those other documents were not also being deleted as part of the same request. When that happens, it leaves behind "broken references" in the database.

Now (in this PR), the endpoint disallows that. If the user tries to delete a document that other documents reference, and those other documents aren't also being deleted as part of the same request, the endpoint returns an HTTP 422 response and does not perform any deletions.

### Details

I updated the `refscan` package version (a dependency of the Runtime) from `0.2.4` to `0.2.5` so that the Runtime could take advantage of a recently-introduced function named `identify_referring_documents`.

Also, to facilitate transitioning to this new "strict" check, I added a boolean flag that a developer can use to switch to a more "lenient" check, where the deletion still happens, but a warning gets logged to the server console (accessible on Rancher). By default, though, this new "strict" mode is enabled.

### Related issue(s)

Fixes #1000 

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I implemented a test that targets this newly-introduced referential integrity checking. I confirmed the test passes locally and on GHA.

### Documentation

- [x] Other (explain below)

I confirmed the API endpoint's user-facing documentation doesn't say anything that this new change would render inaccurate.

### Maintainability

### DIFF
--- a/nmdc_runtime/api/endpoints/queries.py
+++ b/nmdc_runtime/api/endpoints/queries.py
@@ -293,14 +293,23 @@ def _run_mdb_cmd(cmd: Cmd, mdb: MongoDatabase = _mdb) -> CommandResponse:
                             f"Deleting the former will leave behind a broken reference."
                         )
                     else:
+                        # TODO: Consider reporting _all_ would-be-broken references instead of
+                        #       only the _first_ one we encounter. That would make the response
+                        #       more informative to the user in cases where there are multiple
+                        #       such references; but it would also take longer to compute and
+                        #       would increase the response size (consider the case where the
+                        #       user-specified filter matches many, many documents).
                         raise HTTPException(
                             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
                             detail=(
-                                f"Cannot delete the document having 'id'='{target_document_id}' from "
-                                f"the collection '{collection_name}' because it is referenced by "
+                                f"The operation was not performed, because performing it would "
+                                f"have left behind one or more broken references. For example: "
+                                f"The document having 'id'='{target_document_id}' in "
+                                f"the collection '{collection_name}' is referenced by "
                                 f"the document having 'id'='{source_document_id}' in "
                                 f"the collection '{source_collection_name}'. "
-                                f"Deleting it would leave behind a broken reference."
+                                f"Deleting the former would leave behind a broken reference. "
+                                f"Update or delete referring document(s) and try again."
                             ),
                         )
 
--- a/nmdc_runtime/api/endpoints/queries.py
+++ b/nmdc_runtime/api/endpoints/queries.py
@@ -224,15 +224,19 @@ def _run_mdb_cmd(cmd: Cmd, mdb: MongoDatabase = _mdb) -> CommandResponse:
         #       not occur within a transaction. The database may change between the
         #       two events (i.e. there's a race condition).
         #
-        target_document_descriptors = list(mdb.get_collection(collection_name).find(
-            filter={"$or": [spec["filter"] for spec in delete_specs]},
-            projection={"_id": 1, "id": 1, "type": 1},
-        ))
+        target_document_descriptors = list(
+            mdb.get_collection(collection_name).find(
+                filter={"$or": [spec["filter"] for spec in delete_specs]},
+                projection={"_id": 1, "id": 1, "type": 1},
+            )
+        )
         finder = Finder(database=mdb)
 
         # Make a list of all of the (distinct) `_id` values up front, which we can consult
         # when processing _each_ target document descriptor later.
-        target_document_object_ids = list(set([tdd["_id"] for tdd in target_document_descriptors]))
+        target_document_object_ids = list(
+            set([tdd["_id"] for tdd in target_document_descriptors])
+        )
 
         for target_document_descriptor in target_document_descriptors:
             # If the document descriptor lacks the "id" field, we already know that no
@@ -250,7 +254,10 @@ def _run_mdb_cmd(cmd: Cmd, mdb: MongoDatabase = _mdb) -> CommandResponse:
             # to delete, then we know that performing the deletion would leave behind
             # broken references. In that case, we abort with an HTTP 422 error response.
             for referring_document_descriptor in referring_document_descriptors:
-                if referring_document_descriptor["_id"] not in target_document_object_ids:
+                if (
+                    referring_document_descriptor["_id"]
+                    not in target_document_object_ids
+                ):
                     source_document_id = referring_document_descriptor[
                         "source_document_id"
                     ]
--- a/nmdc_runtime/api/endpoints/queries.py
+++ b/nmdc_runtime/api/endpoints/queries.py
@@ -219,10 +219,12 @@ def _run_mdb_cmd(cmd: Cmd, mdb: MongoDatabase = _mdb) -> CommandResponse:
         #       not occur within a transaction. The database may change between the
         #       two events (i.e. there's a race condition).
         #
-        target_document_descriptors = list(mdb[collection_name].find(
-            filter={"$or": [spec["filter"] for spec in delete_specs]},
-            projection={"_id": 1, "id": 1, "type": 1},
-        ))
+        target_document_descriptors = list(
+            mdb[collection_name].find(
+                filter={"$or": [spec["filter"] for spec in delete_specs]},
+                projection={"_id": 1, "id": 1, "type": 1},
+            )
+        )
 
         # Make a set of the `_id` values of the target documents so that (later) we can
         # check whether a given _referring_ document is also one of the _target_ documents
--- a/nmdc_runtime/api/endpoints/queries.py
+++ b/nmdc_runtime/api/endpoints/queries.py
@@ -197,9 +197,10 @@ def _run_mdb_cmd(cmd: Cmd, mdb: MongoDatabase = _mdb) -> CommandResponse:
     logging.info(f"Command type: {type(cmd).__name__}")
     logging.info(f"Cursor ID: {cursor_id}")
 
-    # Initialize a flag that controls whether we will raise an exception
-    # (and abort the operation) or merely log a warning, when we determine
-    # that an operation would leave behind a broken reference(s).
+    # Initialize a flag we can use to control whether we will raise an exception
+    # and abort the operation (i.e. we will be "strict") or merely log a warning
+    # to the console (i.e. we will be "lenient"), when we determine that performing
+    # an operation would leave behind a broken reference(s).
     #
     # Note: We may eventually remove this flag. We are including it now
     #       so that we can easily switch between the two modes, since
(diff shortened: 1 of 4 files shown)


nmdc_runtime/api/endpoints/queries.py
```diff
@@ -31,7 +33,7 @@
     CursorYieldingCommand,
 )
 from nmdc_runtime.api.models.user import get_current_active_user, User
-from nmdc_runtime.util import OverlayDB, validate_json
+from nmdc_runtime.util import OverlayDB, validate_json, get_allowed_references, nmdc_schema_view
```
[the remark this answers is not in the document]
eecavanna: I disagree. The function is used within this module. Closing.

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -201,6 +208,66 @@ def _run_mdb_cmd(cmd: Cmd, mdb: MongoDatabase = _mdb) -> CommandResponse:
...
+        #
+        target_document_descriptors = list(mdb.get_collection(collection_name).find(
+            filter={"$or": [spec["filter"] for spec in delete_specs]},
+            projection={"_id": 1, "id": 1, "type": 1},
+        ))
+        finder = Finder(database=mdb)
+
+        # Make a list of all of the (distinct) `_id` values up front, which we can consult
+        # when processing _each_ target document descriptor later.
+        target_document_object_ids = list(set([tdd["_id"] for tdd in target_document_descriptors]))
```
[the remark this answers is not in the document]
eecavanna: I find `set()` easier to read than `{}` since the latter symbols are overloaded in Python (used for both dicts and sets).

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -201,6 +208,78 @@ def _run_mdb_cmd(cmd: Cmd, mdb: MongoDatabase = _mdb) -> CommandResponse:
...
+
+        # Check whether any of the documents the user wants to delete are referenced
+        # by any documents that are _not_ among those documents. If any of them are,
+        # it means that performing the deletion would leave behind a broken reference(s).
+        #
+        # TODO: Account for the "limit" property of the delete specs.
+        #
+        # TODO: Account for the fact that this validation and the actual deletion do
+        #       not occur within a transaction. The database may change between the
+        #       two events (i.e. there's a race condition).
```
eecavanna, line 220: Still to do/resolve.

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -201,6 +208,84 @@ def _run_mdb_cmd(cmd: Cmd, mdb: MongoDatabase = _mdb) -> CommandResponse:
...
+        # `_id` values are always unique within a collection, so we can use them to
+        # identify recurring target document descriptors.
+        #
+        # Note: We use the `distinct_target_document_object_ids` set for two things:
+        #       1. Here, so we can quickly check whether we have already preserved
+        #          a descriptor for a given document the user wants to delete.
+        #       2. Later, so we can quickly check whether a _referring_ document
+        #          is among the documents the user wants to delete.
+        #
+        distinct_target_document_descriptors = []
```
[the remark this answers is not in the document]
eecavanna: Fixed in commit d04445b0928f7e5b3331730cfdef4c5cbe9d00a1

eecavanna:
Note from ref. int. meeting:
- [ ] Idea: Add a flag in the code that will make it easier to switch this check on/off (in case we find that this drastically inconveniences some of our end users in the short term)

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -190,6 +197,17 @@ def _run_mdb_cmd(cmd: Cmd, mdb: MongoDatabase = _mdb) -> CommandResponse:
...
+    # Initialize a flag that controls whether we will raise an exception
+    # (and abort the operation) or merely log a warning, when we determine
+    # that an operation would leave behind a broken reference(s).
+    #
+    # Note: We may eventually remove this flag. We are including it now
+    #       so that we can easily switch between the two modes, since
+    #       some users have expressed that they may need some time to
+    #       update some client code to work with the more strict mode.
+    #
+    are_broken_references_allowed: bool = False
```
eecavanna, line 209: FYI @aclum, I added a boolean flag that we (a Runtime developer) can toggle in order to switch this newly-introduced real-time referential integrity check (for the "delete" command of the `/queries:run` endpoint) between the following two modes:
- "Strict" (if `False`): This is where the endpoint will return an HTTP 422 response instead of performing any deletions.
- "Lenient" (if `True`): This is where the endpoint will process the deletions normally and will log a warning to the Runtime's console (viewable on Rancher).

tests/test_api/test_endpoints.py
```diff
@@ -1191,6 +1190,89 @@ def test_run_query_update_as_user(api_user_client):
...
+            {
+                "delete": "study_set",
+                "deletes": [
+                    {
+                        "q": {"id": study_b["id"]},
+                        "limit": 0,
+                    }
+                ],
+            },
+        )
```
eecavanna, line 1237: Hi @aclum, here's an example "delete" command where the endpoint will reject the deletion because the document the user wants to delete is still referenced by something.

eecavanna:
I am ready for people to review this PR. I will designate some reviewers.

tests/test_api/test_endpoints.py
```diff
@@ -1191,6 +1190,89 @@ def test_run_query_update_as_user(api_user_client):
         )
 
 
+def test_queries_run_rejects_deletions_that_would_leave_broken_references(
+    api_user_client,
+    fake_studies_and_biosamples_in_mdb,
+):
+    study_a, study_b, bsm_a, bsm_b = fake_studies_and_biosamples_in_mdb
```
eecavanna, line 1197: Here's the graph of relationships between these documents:

```mermaid
flowchart BT
    study_a
    study_b --> |part_of| study_a
    biosample_a --> |associated_studies| study_a
    biosample_b --> |associated_studies| study_b
```

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -201,6 +219,81 @@ def _run_mdb_cmd(cmd: Cmd, mdb: MongoDatabase = _mdb) -> CommandResponse:
...
+                continue
+            referring_document_descriptors = identify_referring_documents(
+                document=target_document_descriptor,  # expects at least "id" and "type"
+                schema_view=nmdc_schema_view(),
+                references=get_allowed_references(),
+                finder=finder,
+            )
+            # If _any_ referring document is _not_ among the documents the user wants
+            # to delete, then we know that performing the deletion would leave behind a
+            # broken reference(s). In that case, we abort with an HTTP 422 error response.
```
eecavanna, line 264: Add note saying we only abort if we are in strict mode.

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -201,6 +219,81 @@ def _run_mdb_cmd(cmd: Cmd, mdb: MongoDatabase = _mdb) -> CommandResponse:
...
+        )
+
+        # Make a set of the `_id` values of the target documents so that (later) we can
+        # check whether a given _referring_ document is also one of the _target_ documents
+        # (i.e. is among the documents the user wants to delete).
+        target_document_object_ids = set(
+            tdd["_id"] for tdd in target_document_descriptors
+        )
+
+        finder = Finder(database=mdb)
```
eecavanna, line 249: Add introductory blurb over this block of code.

tests/test_api/test_endpoints.py
```diff
@@ -1115,7 +1114,7 @@ def _test_run_query_delete_as(client):
     biosample_id = "nmdc:bsm-12-deleteme"
 
     if not mdb.biosample_set.find_one({"id": biosample_id}):
-        mdb.biosample_set.insert_one({"id": biosample_id})
+        mdb.biosample_set.insert_one({"id": biosample_id, "type": "nmdc:Biosample"})
```
dwinston, line 1117: Hmm. Not PR-blocking, but it'd be good to have "partial" `Faker` functionality in the case of e.g. wanting to create a "fake" biosample with a supplied `id` and potentially other fields. That is, "given this partial descriptor, return a 'full' (schema-compliant) biosample document".

dwinston (approved):
Cool, looks good.

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -201,6 +220,90 @@ def _run_mdb_cmd(cmd: Cmd, mdb: MongoDatabase = _mdb) -> CommandResponse:
...
+                    if are_broken_references_allowed:
+                        logging.warning(
+                            f"The document having 'id'='{target_document_id}' in "
+                            f"the collection '{collection_name}' is referenced by "
+                            f"the document having 'id'='{source_document_id}' in "
+                            f"the collection '{source_collection_name}'. "
+                            f"Deleting the former will leave behind a broken reference."
+                        )
+                    else:
+                        raise HTTPException(
```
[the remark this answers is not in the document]
dwinston: To clarify, this is not PR-blocking. Given that we are deploying this behavior using a feature flag, I expect we can communicate with users about what they'd find helpful in an error message.

tests/test_api/test_endpoints.py
eecavanna: The newly-introduced (a few weeks ago) faker module works that way. For each arbitrary kwarg the caller passes to the faker method, the generated fake documents (dicts) will have a key-value pair that matches that kwarg's name and value.

sujaypatil96 (dismissed):
The video of the code walkthrough demo'ing referential integrity assertion on the deletion of documents with/without references looked good to me! very helpful 🙏🏼 excited to test this out in dev once the PR is merged!

tests/test_api/test_endpoints.py
```diff
@@ -730,34 +730,33 @@ def test_get_related_ids_returns_empty_resources_list_for_isolated_subject(
 def fake_studies_and_biosamples_in_mdb():
     # Seed the database with the following interrelated documents:
     # - `study_a`
-    # - `study_b`, which influences (via `part_of`) `study_a`
-    # - `biosample_a`, which influences (via `associated_studies`) `study_a`
-    # - `biosample_b`, which influences (via `associated_studies`) `study_b`
+    # - `study_b` --[part_of]-> `study_a`
+    # - `biosample_a` --[associated_studies]-> `study_a`
+    # - `biosample_b` --[associated_studies]-> `study_b`
```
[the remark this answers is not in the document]
eecavanna: Thanks, @dwinston. I updated this particular comment to use standard Mermaid syntax, so that developers can copy/paste it into GitHub comments or `mermaid.live`.

PR #1008, opened 2025-06-04
Add referential integrity check to `POST /queries:run` API endpoint (for `update` command)

### LINKED ISSUE #1007: Add referential integrity check to `POST /queries:run` API endpoint (for `update` command)
While working on this, also look at how we might be able to get the changesheet endpoints to use some of the same validation under the hood.

eecavanna:
On this branch, we added a real-time referential integrity check to the "update" command of the `POST /queries:run` API endpoint. We also consolidated some class definitions that we came across while working on the former.

### Details

The algorithm I implemented for the real-time referential integrity checking for the "update" command is basically: 

1. Simulate the update and check two things:
   1. Check whether any of the updated documents contain broken references.
   2. Check whether any of the documents that—before the simulated update—_referred_ to any of the updated documents, contain broken references.
4. If either thing is true, abort and respond with an HTTP 422 status code; instead of allowing the remainder of the endpoint's request-handling function to be executed. In other words, don't perform the update.

Under normal circumstances, the API server will now reject commands that it determines would leave behind broken references. However, users can include an `allow_broken_refs=true` query parameter in order to bypass that safety feature (in that case, the server will not reject commands for that reason).

I also updated the `refscan` library from version `0.2.5` to `0.3.0` in order to take advantage of some Python API simplification introduced by version `0.3.0`. Along with updating the library, I also updated existing invocations of its functions, to accommodate breaking changes introduced by the new version of the library.

I also extracted some existing code snippets into reusable functions (named `derive_update_specs` and `derive_delete_specs`), and then documented them and added doctests to them.

I also updated some of the endpoint's error responses (unrelated to referential integrity checking) that, in some situations, did not reflect the underlying error.

### Related issue(s)

Fixes #1007 

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by adding tests that target the new referential integrity checking and ensuring those tests pass locally and on GHA.

### Documentation

- [ ] Other (explain below)

### Maintainability


eecavanna:
To do:
- [ ] Revisit why `main.txt` and `dev.txt` ended up with so many changes (I was expecting only the `refscan` line to have changed)

nmdc_runtime/api/endpoints/lib/helpers.py
```diff
@@ -0,0 +1,201 @@
...
+        # If the document descriptor lacks the "id" field, we already know that no
+        # documents reference it (since they would have to _use_ that "id" value to
+        # do so). So, we don't bother trying to identify documents that reference it.
+        if "id" not in target_document_descriptor:
+            continue
+
+        referring_document_descriptors = identify_referring_documents(
+            document=target_document_descriptor,  # expects at least "id" and "type"
+            schema_view=nmdc_schema_view(),
+            references=get_allowed_references(),
```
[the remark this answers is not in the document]
eecavanna: `get_allowed_references()` is a memoized function.

I don't know about `nmdc_schema_view()` off the top of my head.

Either way, I agree.

nmdc_runtime/api/endpoints/lib/helpers.py
```diff
@@ -0,0 +1,201 @@
...
+from pymongo.database import Database
+from refscan.lib.Finder import Finder
+from refscan.scanner import identify_referring_documents, scan_outgoing_references
+
+from nmdc_runtime.api.models.lib.helpers import derive_update_specs
+from nmdc_runtime.api.models.query import UpdateCommand, UpdateSpecs
+from nmdc_runtime.util import get_allowed_references, nmdc_schema_view
+
+
+def simulate_updates_and_check_references(
```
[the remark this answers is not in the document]
eecavanna: I agree. I plan to either break it up in a future PR, or—if we change direction (e.g. switch to Postgres)—get rid of it completely in a future PR.

eecavanna:
- [ ] TODO: Add an on/off switch for this referential integrity checking, in case team members' downstream clients cannot easily accommodate it yet.

nmdc_runtime/util.py
```diff
@@ -533,13 +533,17 @@ def ensure_unique_id_indexes(mdb: MongoDatabase):
                     raise
 
 
+# TODO: Check whether this class definition is redundant with the definition
+#       of an `UpdateStatement` class in `nmdc_runtime/api/models/query.py`.
 class UpdateStatement(BaseModel):
```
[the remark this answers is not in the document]
eecavanna: I agree with this. I added the `TODO` comment in an attempt to highlight that.

nmdc_runtime/api/models/lib/helpers.py
```diff
@@ -0,0 +1,78 @@
...
+    ... })
+    >>> update_specs = derive_update_specs(update_command)
+    >>> update_specs[0]
+    {'filter': {'color': 'blue'}, 'limit': 0}
+    >>> update_specs[1]
+    {'filter': {'color': 'green'}, 'limit': 1}
+    """
+
+    return [
+        {"filter": update_statement.q, "limit": 0 if update_statement.multi else 1}
```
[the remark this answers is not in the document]
eecavanna: When implementing these helper functions, I opted to preserve the behavior that was already in the `/queries:run` endpoint, under the assumption that the omission of those fields was intentional.

nmdc_runtime/util.py
```diff
@@ -533,13 +533,17 @@ def ensure_unique_id_indexes(mdb: MongoDatabase):
...
     q: dict
     u: dict
     upsert: bool = False
     multi: bool = False
```
eecavanna, line 542: The definition of the `UpdateStatement` class in `nmdc_runtime/api/models/query.py` is (slightly different):

```py
class UpdateStatement(BaseModel):
    q: Document
    u: Document
    upsert: bool = False
    multi: bool = False
    hint: Optional[Dict[str, OneOrMinusOne]] = None
```

nmdc_runtime/util.py
```diff
@@ -533,13 +533,17 @@ def ensure_unique_id_indexes(mdb: MongoDatabase):
...
 
 
+# TODO: Check whether this class definition is redundant with the definition
+#       of a `DeleteStatement` class in `nmdc_runtime/api/models/query.py`.
 class DeleteStatement(BaseModel):
     q: dict
     limit: Annotated[int, Field(ge=0, le=1)] = 1
```
eecavanna, line 549: The definition of the `DeleteStatement` class in `nmdc_runtime/api/models/query.py` is (slightly different):

```py
class DeleteStatement(BaseModel):
    q: Document
    limit: OneOrZero
    hint: Optional[Dict[str, OneOrMinusOne]] = None
```

.env.example
```diff
@@ -80,3 +80,10 @@ IS_RELATED_IDS_ENDPOINT_ENABLED=true
 # Feature flag that can be used to enable/disable the `/scalar` endpoint.
 # Valid values: {`true`, `false`}
 IS_SCALAR_ENABLED=true
+
+# Feature flag that can be used to switch the `/queries:run` endpoint between
+# "lenient" mode (`true`) and "strict" mode (`false`), in terms of referential
+# integrity checking. Unless this is set to `true`, the endpoint will reject
+# operations that would leave behind broken references.
+# Valid values: {`true`, `false`}
+IS_QUERIES_RUN_ENDPOINT_ALLOWING_BROKEN_REFERENCES=false
```
shreddd, line 89: This is a really long variable name - consider renaming since the comment explains it already. maybe ALLOW_QUERIESRUN_LENIENT? Or ALLOW_QUERIESRUN_BROKEN_REFS?

shreddd (commented):
Can we do a code walkthrough on zoom? The code is complex enough that it might be better to go through it togther.

.env.example
eecavanna:
```py
IS_QUERIES_RUN_ALLOWING_BROKEN_REFS
```

nmdc_runtime/util.py
eecavanna: @dwinston will you take a look at this?

nmdc_runtime/api/endpoints/lib/helpers.py
```diff
@@ -0,0 +1,274 @@
...
+        f"the collection '{collection_name}' contains a "
+        f"reference (in its '{source_field_name}' field, "
+        f"referring to the document having id='{target_document_id}') "
+        f"which would be broken."
+    )
+
+
+def simulate_updates_and_check_references(
+    db: Database, update_cmd: UpdateCommand
+) -> List[str]:
```
eecavanna, line 42: The endpoint still only allows someone to update documents in a single collection (per HTTP request), so users are still bound to that. If there are reciprocal links between documents residing in different collections, users may run into a case where they can't perform the update they want to.

eecavanna:
@shreddd and I discussed this PR today. I walked (maybe meandering is more accurate—sorry!) him through the diff.

One thing that we talked about was the fact that the endpoint currently only allows users to manipulate one collection at a time (whether to "update" documents within it, or to "delete" documents within it).

When we introduced the "delete" counterpart of this referential integrity checking a few weeks ago, team members ran into a catch-22 situation involving reciprocal links (prompting us to disable the "delete" counterpart of this checking).

Here, where we are introducing the "update" counterpart, team members may still run into catch-22 situations. That is why this checking is subject to the feature flag. However, this led to @shreddd wondering whether this feature is _not_ a step in the right direction after all.

Another topic that came up was switching to the bulkWrite approach, which allows users to make changes to multiple collections at once. That would be a separate PR.

As a side note, I expressed that I have found "pass-all-this-through-to-Mongo" endpoints like this difficult to maintain (**given that Mongo is not validating things for us**) and would prefer to implement more narrowly-scoped endpoints instead. So, if we were to switch to bulkWrite, maybe we can, for example, separate the "update", "find", "delete", and "aggregated" operations into distinct, more concise endpoints, with simpler user-facing surfaces.

.env.example
shreddd: can this be a query parameter?

eecavanna: I replaced the environment variable with a query parameter in the late commit (i.e. commit https://github.com/microbiomedata/nmdc-runtime/pull/1008/commits/175fbc83fe33e51e4a4c112f946ed3d76b2bbb72).

dwinston:
@eecavanna https://github.com/microbiomedata/nmdc-runtime/pull/1008#discussion_r2197515987 is a bit broader than it's scope, so I thought it wise to draw attention to it via a "top-level" comment.

eecavanna:
> I want to also note, at the top level here, that I think making the limit property for the "delete" statement required constitutes a "breaking change" for HTTP clients.

Oops—sorry 'bout this. I misunderstood the situation as being that the `limit` property was not previously required in user-submitted "delete" statements, and now it was, in which case, I thought this was a user-facing breaking change.

Now, I see that the change only impacted dependents of the now-removed class. The HTTP request validation stage of the `/queries:run` endpoint is not one of those dependents.

Deleting that comment (and this one).

shreddd:
Per donny's comment in meeting - explicitly log that "allow_broken_refs"  flag was used, for admin / obeservability.

eecavanna:
Thanks, @shreddd and @dwinston. Currently (on this branch), the referential integrity checks _always_ happen for "delete" and "update" commands, and the `allow_broken_refs=true` query parameter controls whether—if any violations are found—the server (a) rejects the command and reports the violations to the user, or (b) accepts the command and logs the violations to the server console.

What we talked about adding to that is:
- In case (a), also tell the user that some references were broken (have that info anyway, by that point)
- In case (b), also log that the user used that query parameter

shreddd:
Thanks - let us know when the changes are added in.

eecavanna:
Merging per discussion with @shreddd via Slack. I created a follow-on ticket, [here](https://github.com/microbiomedata/nmdc-runtime/issues/1072), for the outstanding task.

PR #1010, opened 2025-06-04
Configure Dagster run `materialize_alldocs` op only after its dependent ops have finished 

### LINKED ISSUE #1011: Dagster syncs the `alldocs` cache prematurely
These graphs include the `materialize_alldocs` op, which is used to refresh the alldocs collection (cache) to by in sync with the upstream collections.

https://github.com/microbiomedata/nmdc-runtime/blob/4bb75fd36b1168126801b8ec96cfd1776631e256/nmdc_runtime/site/graphs.py#L134-L146

The problem is that, because we are not telling Dagster that the `materialize_alldocs` op _depends_ upon the other ops in the graph, Dagster is running it in parallel with that [chain of] other ops.

The task here is to update the graphs (and whatever else is necessary to update) so that Dagster does not run the `materialize_alldocs` op until the "upstream" ops have finished running.

A fix is in development now.

eecavanna:
On this branch, I updated the two Dagster graphs that include the `materialize_alldocs` op, so that Dagster will not run the `materialize_alldocs` op until the "previous" op has finished running.

### Details

Here's one of those graphs on `main` (notice there is no "dependency" line connected to the `materialize_alldocs` op):

image

Here's the same graph on this branch (notice the new "dependency" line):

image

### Related issue(s)

Fixes #1011 

### Related subsystem(s)

- [ ] Minter
- [x] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by confirming that, on Dagit, the two affected graphs look the way I expect. I also tested the changes by confirming the GHA checks (including tests) all pass.

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/nmdc_runtime/site/ops.py
+++ b/nmdc_runtime/site/ops.py
@@ -1257,7 +1257,7 @@ def _add_related_ids_to_alldocs(
 
 
 # Note: Here, we define a so-called "Nothing dependency," which allows us to (in a graph)
-#       pass an argument the op (in order to specify the order of the ops in the graph)
+#       pass an argument to the op (in order to specify the order of the ops in the graph)
 #       while also telling Dagster that this op doesn't need the _value_ of that argument.
 #       This is the approach shown on: https://docs.dagster.io/api/dagster/types#dagster.Nothing
 #       Reference: https://docs.dagster.io/guides/build/ops/graphs#defining-nothing-dependencies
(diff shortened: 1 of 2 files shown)


nmdc_runtime/site/ops.py
```diff
@@ -38,7 +38,8 @@
     Optional,
     Field,
     Permissive,
-    Bool,
```
eecavanna, line 41: This was an unused `import`. I removed it.

sujaypatil96 (dismissed):
Awesome!! thanks for these changes that ensure that the ops in the job run in the order that we expect them to be run @eecavanna 🚀

PR #1019, opened 2025-06-17
Fix incorrect `/queries:run` "update" command example (Swagger UI)

### LINKED ISSUE #1015: Swagger UI shows incorrect example for `/queries:run` "update" command
The documentation says:

Image

In reality, the update will only affect 0-1 documents. By default, there is a "limit" of 1 applied.

https://github.com/microbiomedata/nmdc-runtime/blob/46c150f4c3ada09c27d0411388f7e2b7de6cf0db/nmdc_runtime/api/endpoints/queries.py#L350-L352

https://github.com/microbiomedata/nmdc-runtime/blob/46c150f4c3ada09c27d0411388f7e2b7de6cf0db/nmdc_runtime/api/models/query.py#L174-L181

The user can update an unlimited number of documents by setting `multi` to `true` in the command. Example:

```json
{
  "update": "study_set",
  "updates": [{"q": {"...": "..."}, "u": {"$set": {"...": "..."}}, "multi": true}]
}
```

eecavanna:
On this branch, I updated an example request payload shown on Swagger UI for the `/queries:run` API endpoint, so that it is correct. On `main`, it is currently incorrect (it is missing the `"multi": true` property).

### Details

I also added a second example "update" command. Now, one example illustrates the updating of a single matching document, and another one illustrates the updating of all matching documents.

### Related issue(s)

Fixes #1015 

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I did not test these changes. I made the changes using GitHub's `github.dev` editor.

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/nmdc_runtime/api/endpoints/queries.py
+++ b/nmdc_runtime/api/endpoints/queries.py
@@ -122,21 +122,25 @@ def run_query(
     }
     ```
 
-    Rename the first biosample having a given `id`.
+    Rename the first\* embargoed biosample.
     ```
     {
       "update": "biosample_set",
-      "updates": [{"q": {"id": "A_BIOSAMPLE_ID"}, "u": {"$set": {"name": "A_NEW_NAME"}}}]
+      "updates": [{"q": {"embargoed": true}, "u": {"$set": {"name": "A_NEW_NAME"}}}]
     }
     ```
 
-    Rename all biosamples that are embargoed.
+    \*<small>Updates at most 1 matching document, since `"multi": true` is not present.</small>
+
+    Rename all\* embargoed biosamples.
     ```
     {
       "update": "biosample_set",
       "updates": [{"q": {"embargoed": true}, "u": {"$set": {"name": "A_NEW_NAME"}}, "multi": true}]
     }
     ```
+
+    \*<small>Updates all matching documents, since `"multi": true` is present.</small>
 
     Get all\* biosamples, sorted by the number of studies associated with them (greatest to least).
     ```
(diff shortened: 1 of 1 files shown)


hesspnnl (commented):
I find the wording still a little bit confusing for documentation. If the first example is strictly 0-1 record changes, I feel the comment should reflect that. Something like "Rename strictly 0 or 1 records that match the given id." The second example's comment makes sense to me.

hesspnnl:
> Thanks, @hesspnnl, for giving me that feedback. I updated the payloads to be more consistent with one another (so the presence of `"multi": true` is the _only_ difference between them). I also updated the explanations to be more explicit about what `"multi": true` does. What do you think of this new version?
> 
> image
> 
> From my perspective, there's still some ambiguity: How do we determine which document is the _first_ one? In reality, it's the first one that Mongo happens to encounter when it's processing the query. For reference, the same language is used in an example above these ones (i.e. "Delete the _first_ biosample having a given `id`"). Maybe saying "the first embargoed biosample" prompts the question about ordering, whereas saying "one embargoed biosample" or "up to one embargoed biosample" does not. I don't want the information to be inaccurate, and also don't want people to find the wording awkward.

Hmm interesting thoughts. Is there a rhyme or reason to which record gets changed in the embargoed scenario? Is it LIFO, FIFO, something else, or just random? I think if it is random (or if we want the users of the endpoint to _think_ it is random) then I think what you changed it to is sufficient, clear, and is definitely an improvement from the last iteration.

eecavanna:
Thanks, @hesspnnl.

> or if we want the users of the endpoint to think it is random

Given the way the endpoint has been implemented, I want the user to think it is random (I think of it as being random). The endpoint doesn't do any sorting (e.g. sorting by `id`, for example) before performing the update.

PR #1092, opened 2025-07-23
Update `related_resources` endpoint to account for multiple `DataGeneration`s informing a `WorkflowExecution`

### LINKED ISSUE #1083: Confirm correctness of specific change within "related_resources" endpoint
In commit f61cf127, I (@eecavanna) attempted to change the `find_related_objects_for_workflow_execution` function to reflect the fact that, in the new version of the schema, the `was_informed_by` slot was changing from single-valued to multivalued. All tests passed, but, given my only-superficial familiarity with the algorithm, I'm concerned that this is insufficient (e.g. maybe we have to find _all_ `DataGeneration`s—instead of only one—whose `id` value is in the `was_informed_by` list).

The relevant change to the function was:

```diff
        # Look for a DataGeneration in the `alldocs` collection.
        # We'll use that DataGeneration to get to related Biosamples.
-       dg_doc = mdb.alldocs.find_one({"id": was_informed_by})
+       dg_doc = mdb.alldocs.find_one({"id": {"$in": was_informed_by}})
```

This ticket represents the task of confirming that this is, indeed, a sufficient change to accommodate that schema change.

Note that commit f61cf127 included some other changes to the same function. I am not concerned about those.

The next production release is scheduled for https://github.com/microbiomedata/issues/issues/1263

CC: @sujaypatil96, @aclum

eecavanna:
On this branch, I updated the API endpoint to return `Biosample`s related to a `WorkflowExecution` through _any_ `DataGeneration`s that inform that `WorkflowExecution`, as opposed to only one of the `DataGeneration`s that informs it.

### Details

To do that, I took a TDD approach: I first implemented a test demonstrating that (on `main`), when a `WorkflowExecution` "was informed by" two `DataGeneration`s, the API response only included the `Biosample` associated with _one_ of those `DataGeneration`s. Then, I updated the endpoint so that its response includes both `Biosample`s.

I also updated the endpoint's description to reflect the fact that its response payload does _not_ include `DataGeneration`s.

### Related issue(s)

Fixes #1083 

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by implementing a test demonstrating the bug. The test failed before I made changes to the endpoint, then passed after I made changes to the endpoint. All other tests still pass.

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/nmdc_runtime/api/endpoints/find.py
+++ b/nmdc_runtime/api/endpoints/find.py
@@ -410,10 +410,9 @@ def find_planned_process_by_id(
         "This endpoint returns a JSON object that contains "
         "(a) the specified `WorkflowExecution`, "
         "(b) all the `DataObject`s that are inputs to — or outputs from — the specified `WorkflowExecution`, "
-        "(c) all the `DataGeneration`s that generated those `DataObject`s, "
-        "(d) all the `Biosample`s that were inputs to those `DataGeneration`s, "
-        "(e) all the `Study`s with which those `Biosample`s are associated, and "
-        "(f) all the other `WorkflowExecution`s that are part of the same processing pipeline "
+        "(c) all the `Biosample`s that were inputs to those `DataGeneration`s, "
+        "(d) all the `Study`s with which those `Biosample`s are associated, and "
+        "(e) all the other `WorkflowExecution`s that are part of the same processing pipeline "
         "as the specified `WorkflowExecution`."
         "<br /><br />"  # newlines
         "**Note:** The data returned by this API endpoint can be up to 24 hours out of date "
--- a/nmdc_runtime/api/endpoints/find.py
+++ b/nmdc_runtime/api/endpoints/find.py
@@ -655,9 +655,7 @@ def find_related_objects_for_workflow_execution(
         # get all `Biosample`s and `Study`s associated with each of those `DataGeneration`s.
         dg_docs = mdb.alldocs.find({"id": {"$in": was_informed_by}})
         for dg_doc in dg_docs:
-            if any(
-                t in dg_descendants for t in dg_doc.get("_type_and_ancestors", [])
-            ):
+            if any(t in dg_descendants for t in dg_doc.get("_type_and_ancestors", [])):
                 # Get Biosamples from the DataGeneration's `has_input` field by recursively walking up the chain.
                 # While we recursively walk up the chain, we'll add those Biosamples to our list of Biosamples.
                 for input_id in dg_doc.get("has_input", []):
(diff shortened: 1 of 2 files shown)


nmdc_runtime/api/endpoints/find.py
```diff
@@ -652,28 +651,27 @@ def process_id(current_id):
...
-            # add the Studies associated with those Biosamples to our list of Studies.
-            if not dg_doc.get("associated_studies") and len(biosamples) > 0:
-                for bs in biosamples:
-                    for study_id in bs.get("associated_studies", []):
-                        add_study(study_id)
+        # Get all `DataGeneration`s that informed the user-specified `WorkflowExecution`, then
+        # get all `Biosample`s and `Study`s associated with each of those `DataGeneration`s.
+        dg_docs = mdb.alldocs.find({"id": {"$in": was_informed_by}})
+        for dg_doc in dg_docs:
+            if any(t in dg_descendants for t in dg_doc.get("_type_and_ancestors", [])):
```
eecavanna, line 658: Note to reviewers: These are the only lines in the endpoint, that I edited beyond their indentation.

tests/test_api/test_endpoints.py
```diff
@@ -2056,6 +2056,90 @@ def test_find_related_resources_for_workflow_execution__returns_related_workflow
...
+        data_generation_a --> |associated_studies| study
+        data_generation_b --> |associated_studies| study
+        data_generation_a --> |has_input| biosample_a
+        data_generation_b --> |has_input| biosample_b
+        data_generation_a --> |has_output| data_object_a
+        data_generation_b --> |has_output| data_object_b
+        workflow_execution --> |has_input| data_object_a
+        workflow_execution --> |has_input| data_object_b
+        workflow_execution --> |was_informed_by| data_generation_a
+        workflow_execution --> |was_informed_by| data_generation_b
```
eecavanna, line 2082: Per today's infrastructure meeting:

If a given `WorkflowExecution` is informed by two `DataGeneration`s (e.g. because the researchers decided the first one was insufficient), and those two DataGenerations have the _same_ `Biosample` as input, team members expect this endpoint to still return that `Biosample`.

The endpoint (on `main`) already does that (since it would find that `Biosample` via the one `DataGeneration` it happens to process). This PR makes the algorithm even more general (i.e. it makes it so that, if those two `DataGeneration`s each had a _distinct_ `Biosample` as an input, the endpoint would return _both_ of those `Biosample`s, instead of only one).

eecavanna:
I reviewed this in real time with @shreddd and @sujaypatil96 via Zoom on Wednesday (during the "Runtime" squad meeting). I took an action item to check in with someone more familiar with real-world use cases (e.g. @aclum), which I did during today's infrastructure meeting (see notes from that check-in [here](https://github.com/microbiomedata/nmdc-runtime/pull/1092#discussion_r2229477064)).

sujaypatil96 (approved):
Thanks for checking in about the practical/real world use case/scenario on call today @eecavanna. The changes in this PR look good to me! 🚀

PR #1096, opened 2025-07-24
Implement job deletion in cascading delete

### LINKED ISSUE #1094: additional cascading delete support
When deleting a workflow execution ID from mongo the record should also delete the corresponding record in the jobs collection so that the ID is not reused. @dwinston implemented an unrelease endpoint but that is not appropriate if the record has already been recorded in mongo (b/c it will reuse an ID that has already been deleted).

This does not pertain to mass spec based workflows since nmdc_automation is not used to schedule that work) so excludes [MetabolomicsAnalysis](https://microbiomedata.github.io/nmdc-schema/MetabolomicsAnalysis/), https://microbiomedata.github.io/nmdc-schema/NomAnalysis/, and https://microbiomedata.github.io/nmdc-schema/MetaproteomicsAnalysis/

Example 

when deleting `nmdc:wfrbt-11-46mzv464.1` from `workflow_exeuction_set` cascading delete should apply the following filter to the `jobs` collection`{'config.activity_id':'nmdc:wfrbt-11-46mzv464.1'}` and delete the record, in this case `id` `nmdc:07139eaa-5dbb-11f0-a17d-b61a4536de2d`

shreddd:
Code to delete jobs with associated with a workflow when doing a cascading delete.

- [X] Modify `delete_workflow_execution` to account for jobs
- [X] Add function to implement non schema delete operations.
- [X] Update tests

Fixes #1094


nmdc_runtime/api/endpoints/queries.py
```diff
@@ -607,3 +607,73 @@ def _run_mdb_cmd(
...
+    cmd_response_raw: dict = mdb.command(
+        bson.json_util.loads(json.dumps(cmd.model_dump(exclude_unset=True)))
+    )
+
+    # Create the command response object (assume DeleteCommandResponse type)
+    cmd_response = DeleteCommandResponse(**cmd_response_raw)
+
+    # Check if the command was successful
+    if not cmd_response.ok:
+        return cmd_response
```
eecavanna, line 659: I'm a little confused about this.

If the command response is not OK, we return the command response.

If the command response _is_ OK, we check for some other kinds of errors and — if we don't find them — we again return the command response.

Maybe this is confusion I'd have about `_run_mdb_cmd` also if I were to see all this logic in one place (that function is so "tall" that I don't know that I've seen all of these branches in one place before).

nmdc_runtime/api/endpoints/queries.py
```diff
@@ -607,3 +607,73 @@ def _run_mdb_cmd(
...
 
     return cmd_response
+
+
+def _run_delete_nonschema(
+    cmd: DeleteCommand, mdb: MongoDatabase = _mdb
+) -> DeleteCommandResponse:
+    """
+    Performs deletion operations similar to _run_mdb_cmd but skips the target_document
+    descriptor code that handles referential integrity checking.
```
eecavanna, line 617: I changed "similar" (adjective) to "similarly" (adverb) to emphasize that the function performs the deletions in a similar way to the way `_run_mdb_cmd` does.

```suggestion
    Performs deletion operations similarly to _run_mdb_cmd but skips the target_document
    descriptor code that handles referential integrity checking.
```

nmdc_runtime/api/endpoints/workflows.py
```diff
@@ -129,11 +133,12 @@ async def post_workflow_execution(
     "/workflows/workflow_executions/{workflow_execution_id}",
     response_model=DeleteResponse,
     description="Delete a workflow execution and its downstream workflow executions, data objects, "
-    "and functional annotation aggregation members.\n\n"
+    "functional annotation aggregation members, and related job records.\n\n"
```
eecavanna, line 136: Thanks for updating the docs!

nmdc_runtime/api/endpoints/workflows.py
```diff
@@ -306,7 +320,10 @@ def recursive_delete_workflow_execution(wfe_id: str) -> None:
                 f"Executing cascading delete command for {collection_name} - you may temporarily encounter broken references."
             )
             # Execute the delete command
-            response = _run_mdb_cmd(delete_cmd, mdb, allow_broken_refs=True)
+            if collection_name == "jobs":
```
eecavanna, line 323: Hard-coding `jobs` makes sense here, given our situation. Since the function being called can handle deleting documents from arbitrary collections, I wanted to take this opportunity to say that we can get a list of all schema-described collections via a `SchemaView` (I suspect we are already doing it somewhere). Anyway, hard-coding `"jobs"`, in this case, makes sense to me.

eecavanna (approved):
Looks great! I especially like the test that was augmented to include jobs. I left a few comments, none of which I consider to be deal breakers.

I am confused about the logic around checking `command_response.ok`. I added a comment containing my "out-loud thoughts" (in the "thinking out loud" sense).

eecavanna:
I finished reviewing/approving this PR. I will invite Copilot to do a review also.

nmdc_runtime/api/endpoints/workflows.py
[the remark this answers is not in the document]
eecavanna: Closing. Redundant with https://github.com/microbiomedata/nmdc-runtime/pull/1096#discussion_r2227036461

eecavanna:
Heads up, @pkalita-lbl — we are planning to merge this PR branch into `main` today, even though the release is on Monday.

Rationale:
- The code change closes a gap in the endpoint's current functionality, which is currently closed via manual intervention by a Mongo admin (e.g. me) and an Slack-based approval process involving @aclum and whoever used the endpoint (so, I think it'll save a few of us time and context switching)
- Its scope is small, IMO (it affects only one endpoint), and I think the amount by which its inclusion would increase demand on "user acceptance testers" over the next few days is small, and I think reverting it would be straightforward for a developer (e.g. me)

Issue #1101, opened 2025-07-25
API endpoint `DELETE /workflows/workflow_executions/{id}` sometimes times out


eecavanna:
The cascading deletion endpoint is timing out in the development environment when being used to delete the `WorkflowExecution` whose `id` is `nmdc:wfmtan-11-x3en2872.1`. This was reported by @aclum in the following issue comment:
- https://github.com/microbiomedata/nmdc-runtime/issues/1094#issuecomment-3115340962

I'm creating this ticket so there is a ticket specifically about the performance (i.e. response time) of that endpoint. Once I create this ticket, I'll re-close #1904, which was originally about new functionality (i.e. updating the endpoint to delete `jobs` documents also).

CC: @shreddd

eecavanna:
The endpoint does issue a query that involves looking for `jobs` documents based upon the value in their `config.activity_id` field.

https://github.com/microbiomedata/nmdc-runtime/blob/0c2c0f00183ff7b369ec3024c0fe170c5ad41215/nmdc_runtime/api/endpoints/workflows.py#L270-L273

That field is not currently being indexed.

eecavanna:
I created a PR that creates the index. I am ready for the PR to be reviewed, approved, and merged.

https://github.com/microbiomedata/nmdc-runtime/pull/1102

eecavanna:
Thanks, @aclum. I'll contact you via Slack momentarily to ask about me experimenting with that same `WorkflowExecution` `id` in the development environment.

eecavanna:
I looked into this. I learned three things:
1. As of 10:30am PT, there is no document having that `id` in the `workflow_execution_set` collection.
   ```js
   db["workflow_execution_set"].countDocuments({id: "nmdc:wfmgan-11-c861q114.1"});
   // -> 0
   ```
2. The logs of the `runtime-api` container on Rancher contain about 25 (twenty five) lines matching (with various `data_object_set` document `id` values):
   ```console
   The document having 'id'='nmdc:wfmgan-11-c861q114.1' in the collection 'workflow_execution_set' is referenced by the document having 'id'='...' in the collection 'data_object_set'. Deleting the former will leave behind a broken reference.
   ```
3. The logs of the `runtime-api` container on Rancher contain about **27,000 (twenty-seven thousand)** lines matching (with empty `functional_annotation_agg` document `id` values):
   ```console
[...]

PR #1102, opened 2025-07-25
Update boot code to ensure `jobs` collection has index on `config.activity_id` field

eecavanna:
On this branch, I updated the standard index-creation code that runs as part of the Runtime's boot-up sequence, so that it creates (if one doesn't already exist) and index on the `config.activity_id` field of the `jobs` collection.

### Details

This is all in an attempt to speed up some queries that are run within the `DELETE /workflows/workflow_executions/{workflow_execution_id}` endpoint.

I also added a comment to a separate function, in an attempt to make it easier for people to understand that that function affects both schema-described collections and non-schema-described collections.

### Related issue(s)

At least mitigates, if not fixes, issue #1101

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by manually creating the index in my local development environment and using...

```js
db.jobs.find({"config.activity_id": "nmdc:mgrqc0xxxx.1"}).explain("executionStats")["executionStats"]["executionTimeMillis"];
```

...to check how long the execution time was when (a) the index was active and (b) when the index was hidden. The numbers bounced a round a bit, but I came away with the impression that creating the index will result in around a 10x speed boost to that query.

I will also confirm that all the GHA tests still pass.

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/nmdc_runtime/api/models/util.py
+++ b/nmdc_runtime/api/models/util.py
@@ -252,4 +252,8 @@ entity_attributes_to_index = {
         "has_input",
         "has_output",
     },
+    # Note: The `jobs` collection is not described by the NMDC schema.
+    "jobs": {
+        "config.activity_id",
+    },
 }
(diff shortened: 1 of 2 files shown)


eecavanna:
I want to merge this in before [Monday's release](https://github.com/microbiomedata/issues/issues/1263). The endpoint that this speeds up will be introduced into the production API as part of that release, and I want to have this "fix" be in place when the endpoint gets introduced.

Issue #1107, opened 2025-07-29
Delete top-level `components` directory


eecavanna:
The top-level `components` directory contains 6 files:

```
1. components/nmdc_runtime/workflow/__init__.py
2. components/nmdc_runtime/workflow/spec.py
3. components/nmdc_runtime/workflow_execution_activity/__init__.py
4. components/nmdc_runtime/workflow_execution_activity/core.py
5. components/nmdc_runtime/workflow_execution_activity/spec.py
6. components/nmdc_runtime/workflow_execution_activity/store.py
```

I think all of these files are the remnants of a former team member's attempt at implementing some Runtime functionality using a component-based architecture.

These files are not referenced by anything else in the repository. Indeed, when they are gone, all current tests pass.

CC: @dwinston

PR #1108, opened 2025-07-29
Configure standard test process to include running doctests

### LINKED ISSUE #883: Configure test runners to also run doctests
#### Background

The "Python application" GHA workflow contains a job named "build" which runs tests via:

https://github.com/microbiomedata/nmdc-runtime/blob/89c7d03912d5f8aa16282e4988f382479893a7bf/.github/workflows/python-app.yml#L54-L57

The Make targets referenced above are defined like this:

https://github.com/microbiomedata/nmdc-runtime/blob/89c7d03912d5f8aa16282e4988f382479893a7bf/Makefile#L41-L57

#### Tasks

- [ ] The task here is to confirm whether the GHA workflow runs doctests; and...
- [ ] ...if it does not, make it so it does. In practice, this may involve editing the GHA workflow file, and/or the Makefile, and/or some pytest configuration file, and/or the `docker-compose.test.yml` file.

#### Reference

Here are some example doctests in the repo:

https://github.com/microbiomedata/nmdc-runtime/blob/89c7d03912d5f8aa16282e4988f382479893a7bf/nmdc_runtime/minter/config.py#L17-L34

https://github.com/microbiomedata/nmdc-runtime/blob/89c7d03912d5f8aa16282e4988f382479893a7bf/nmdc_runtime/util.py#L80-L103

https://github.com/microbiomedata/nmdc-runtime/blob/89c7d03912d5f8aa16282e4988f382479893a7bf/nmdc_runtime/api/endpoints/util.py#L157-L167

eecavanna:
On this branch, I updated the `run-test` Make target to run doctests (in addition to the tests it already runs). I also fixed some broken doctests.

### Details

There were two directories that, when not ignored by pytest, pytest would complain about when running doctests. Since those directories have no tests or doctests defined within them, I opted to ignore them for now, and add a `TODO` comment about addressing the root cause.

In the case of the `components` directory, I expect us to delete that directory (per https://github.com/microbiomedata/nmdc-runtime/issues/1107).

In the case of the `nmdc_runtime/api/db/mongo.py` and `nmdc_runtime/api/db/s3.py` files, I don't understand why these ones were causing a problem. Here's the error from pytest (the error was pretty much the same for each of those two files):

```py
ImportError while importing test module '/code/nmdc_runtime/api/db/mongo.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/local/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
E   ModuleNotFoundError: No module named 'db.mongo'
```

### Related issue(s)

Fixes #883 

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [x] Other

### Testing

I tested these changes by confirming all doctests — and regular tests — pass.

### Documentation

- [ ] Other (explain below)

### Maintainability


Makefile
```diff
@@ -77,10 +77,19 @@ reset-db-test:
...
+#
 run-test:
 	docker compose --file docker-compose.test.yml exec -it test \
-		./.docker/wait-for-it.sh fastapi:8000 --strict --timeout=300 -- pytest --cov=nmdc_runtime \
-		$(ARGS)
+		./.docker/wait-for-it.sh fastapi:8000 --strict --timeout=300 -- \
+			pytest --cov=nmdc_runtime \
+				   --doctest-modules \
+				   --ignore=nmdc_runtime/api/db/ \
+				   --ignore=components/ \
```
eecavanna, line 91: I added these three CLI options to the command. All other changes here were either formatting or commentary.

nmdc_runtime/util.py
```diff
@@ -324,9 +324,9 @@ def find_one(k_v: dict, entities: Iterable[dict]):
     """Find the first entity with key-value pair k_v, if any?
 
     >>> find_one({"id": "foo"}, [{"id": "foo"}])
+    {'id': 'foo'}
+    >>> find_one({"id": "foo"}, [{"id": "bar"}]) is None
     True
-    >>> find_one({"id": "foo"}, [{"id": "bar"}])
-    False
```
eecavanna, line 329: These tests were failing as they were written. I updated the "expectations" to match reality. I suspect the author wrote the tests and then changed the behavior of the function without updating the tests (prior to this PR, they were not being run as part of the automated test suite).

eecavanna:
Thanks, @dwinston! I will remove the corresponding comments from the Makefile now.

dwinston (approved):
LGTM

eecavanna:
Thanks! Once the GHA tests pass (which now includes doctests), I will merge this into `main`.

eecavanna:
I opened up a follow-on ticket here (https://github.com/microbiomedata/nmdc-runtime/issues/1110) about the various _warnings_ that appear in the pytest output. Although I don't know how impactful the warned-about issues are to the operation of the Runtime, I do like the idea of cleaning up the pytest output (less scrolling when looking for the main information).

PR #1145, opened 2025-08-14
Migrate from `MongoClient` to `AsyncMongoClient`

### LINKED ISSUE #787: Explore migrating from pymongo to motor as an async python libary for mongodb
Given that there is now "official" guidance from MongoDB Inc. on using their official async python library for mongodb, i.e. [motor](https://pypi.org/project/motor/), with FastAPI, explore pros/cons of migrating from using the older, more-stable-but-not-async [pymongo](https://pypi.org/project/pymongo/) library in order to take full performance advantage of FastAPI's async support.

The implication of this would be that a single-container deployment of the runtime would: 
- support a higher rate of requests (avoid having to run multiple containers)
- in turn, potentially help avoid having to consider race conditions wrt multiple runtime containers accessing the same backing mongodb instance

**Goal**: take some time to compare both approaches to better inform a path forward

eecavanna:
🚧 **THIS BRANCH IS UNDER CONSTRUCTION** 🚧 

On this branch, I migrated the Runtime from using `MongoClient` to use `AsyncMongoClient`. That involved updating the `pymongo` version and doing the things listed in the [official migration guide](https://www.mongodb.com/docs/languages/python/pymongo-driver/current/reference/migration/#migrate-from-pymongo).

### Details

The summary is written in the past tense, but it isn't true yet. Everything here is in progress.

### Related issue(s)

Fixes #787 

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by...

### Documentation

- [ ] Other (explain below)

### Maintainability

- [ ] The PR title is in the imperative mood (e.g. "Do X") and not the declarative mood (e.g. "Does X" or "Did X")

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -23,6 +23,10 @@ certifi==2025.4.26
     # via
     #   -c requirements/main.txt
     #   requests
+cffi==1.17.1
+    # via
+    #   -c requirements/main.txt
+    #   cryptography
 charset-normalizer==3.4.2
     # via
     #   -c requirements/main.txt
@@ -36,6 +40,10 @@ coverage==7.8.0
     # via
     #   -r requirements/dev.in
     #   pytest-cov
+cryptography==44.0.3
+    # via
+    #   -c requirements/main.txt
+    #   secretstorage
 docutils==0.21.2
     # via readme-renderer
 exceptiongroup==1.3.0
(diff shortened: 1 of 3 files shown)


eecavanna:
I did some performance profiling (of individual requests) using `pyinstrument` and found that the `/nmdcschema/{collection_name}` endpoint is spending most of its time, by far, generating the pagination token.

image

As of now, on this branch, that part of the endpoint still uses the _synchronous_ Mongo client (indeed, it's the _only_ part that uses it). I will focus on migrating that to use the _asynchronous_ one (I did it last night, but at the expense of breaking Dagster in my local environment—I'll do it differently today).

PR #1166, opened 2025-08-23
Speed up `list_resources` function for large collections

### LINKED ISSUE #1164: Replace slow `count_documents` call with limited aggregation pipeline (performance)
I came across this `count_documents` invocation last night and added the `TODO` comment above it. This is in a function called `list_resources`, which is used by several endpoints.

https://github.com/microbiomedata/nmdc-runtime/blob/722c3f6073312dc141abc9349efe03bbefadfd24/nmdc_runtime/api/endpoints/util.py#L119-L123

We may be able to apply the technique shown in https://stackoverflow.com/a/67503437.

The idea is that, with `count_documents`, we are counting _all_ the (matching) documents, which may be more counting than is necessary here.

To do:
- Compare performance of `count_documents` and the aggregation pipeline
- If aggregation pipeline is "much" faster (enough to justify complicating the code), use it instead
- Update the `TODO` comment accordingly

eecavanna:
On this branch, I updated the `list_resources` function so that, instead of counting _all_ matching documents in a collection, it only counts _the minimum number necessary_ in order to do its job.

Here's an analogy:

image

I have a yard with lots of different types of trees in it. I want to know whether my yard has **more than 3** oak trees. I can either:
1. (the old way) Ask someone to **count all** the oak trees in my yard, and tell me that number (I'll compare it with 3); or
2. (the new way) Ask someone to **count up to 4** oak trees in my yard, and tell me that number (I'll compare it with 3)

### Details

As a result, the `GET /nmdc_schema/{collection_name}` endpoint (and other endpoints and functions that use the `list_resources` function under the hood) is faster. See load testing results and performance profiling results below for information about "how much faster."

#### Load testing results

| Before<br/>(i.e. commit 722c3f6073312dc141abc9349efe03bbefadfd24, tip of `main` branch) | After<br/>(i.e. commit https://github.com/microbiomedata/nmdc-runtime/pull/1166/commits/7789b00e96742d4befe3a0b978efc0bf99da0c23, near tip of PR branch) |
| --- | --- |
| image | image |
| Median response time for<br/>`GET /nmdcschema/functional_annotation_agg`<br/>under the test load: 76 seconds | Median response time for<br/>`GET /nmdcschema/functional_annotation_agg`<br/>under the test load: 300 milliseconds (~250x faster) | 

<details>
<summary>Show/hide load testing notes</summary>

Here's the command I used to launch the load test(er):

```sh
docker run --rm -it \
    --publish 8089:8089 \
    --volume ${PWD}/util/load_testing:/mnt/locust \
    --network nmdc-runtime-dev_default \
    locustio/locust \
      --locustfile /mnt/locust/locustfile.py \
      --host http://fastapi:8000 --autostart --users 25 --spawn-rate 5 -t 30s User --stop-timeout 180s
```

Notes:
- In the "Before" case, I used the `locustfile.py` file as it existed in the "After" case (i.e. same load configuration in both cases).
- I included `--stop-timeout 180s` in the command because—without it—in the "Before" case, the HTTP responses to the `GET /nmdcschema/functional_annotation_agg` requests were taking so long to come back (about 2 minutes, although that's dependent upon my local environment) that the load tester client would not acknowledge them at that point anymore. When using this 3-minute cleanup period, the load tester client does acknowledge them.

</details>

#### Performance profiling results

Here's how this change impacts the `GET /nmdcschema/functional_annotation_agg` endpoint's performance:

| Before<br/>(i.e. commit 722c3f6073312dc141abc9349efe03bbefadfd24, tip of `main` branch) | After<br/>(i.e. commit https://github.com/microbiomedata/nmdc-runtime/pull/1166/commits/7789b00e96742d4befe3a0b978efc0bf99da0c23, near tip of PR branch) |
| --- | --- |
| image | image |
| Response time: ~20 seconds | Response time: ~20 milliseconds (~1000x faster) |

<details>
<summary>Show/hide performance profiling notes</summary>

Notes:
- In both the "Before" and "After" cases, I prepended `async` to the definition of the endpoint function.
  ```diff
    @router.get(
        "/nmdcschema/{collection_name}",
        response_model=ListResponse[Doc],
        response_model_exclude_unset=True,
    )
  - def list_from_collection(
  + async def list_from_collection(
  ```
- The performance profiling was performed when the Runtime was not receiving any requests _other than_ the one being profiled.

</details>

Also on this branch:
- I added a check for `max_page_size` being something other than an integer, since its type hints say it can be `int | None`.
- I refactored some code in an attempt to make the conditions under which pagination will be performed more apparent to maintainers of this codebase.
- I refined the load testing instructions and related configuration file, based upon my experience measuring the performance of the `GET /nmdcschema/functional_annotation_agg` endpoint. I also configured Uvicorn's file watcher to ignore those files, since they aren't part of the FastAPI application.

### Related issue(s)

Fixes #1164 

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

The test I added, which targets the newly-added helper function, passes locally and on GHA. I also performed load testing and performance profiling (shown above).

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/nmdc_runtime/api/endpoints/util.py
+++ b/nmdc_runtime/api/endpoints/util.py
@@ -138,9 +138,9 @@ def list_resources(req: ListRequest, mdb: MongoDatabase, collection_name: str):
     elif max_page_size < 1:
         will_paginate = False
     elif not does_num_matching_docs_exceed_threshold(
-            collection=mdb[collection_name], filter_=filter_, threshold=max_page_size
-        ):
-            will_paginate = False
+        collection=mdb[collection_name], filter_=filter_, threshold=max_page_size
+    ):
+        will_paginate = False
 
     if not will_paginate:
         rv = {
--- a/nmdc_runtime/api/endpoints/util.py
+++ b/nmdc_runtime/api/endpoints/util.py
@@ -128,7 +128,7 @@ def list_resources(req: ListRequest, mdb: MongoDatabase, collection_name: str):
         id_field = (
             "_id"  # currently expected for `functional_annotation_agg` collection
         )
-    limit = req.max_page_size
+    max_page_size = req.max_page_size
     filter_ = json_util.loads(check_filter(req.filter)) if req.filter else {}
     projection = (
         list(set(comma_separated_values(req.projection)) | {id_field})
@@ -154,15 +154,15 @@ def list_resources(req: ListRequest, mdb: MongoDatabase, collection_name: str):
     # Determine whether we will paginate the results.
     #
     # We will paginate them unless either of the following is true:
-    # - the `limit` is not a positive integer
-    # - the number of documents matching the filter is no larger than `limit`
+    # - the `max_page_size` is not a positive integer
+    # - the number of documents matching the filter is no larger than `max_page_size`
     #
     will_paginate = True
-    if (not isinstance(limit, int)) or (limit < 1):
+    if (not isinstance(max_page_size, int)) or (max_page_size < 1):
         will_paginate = False
     else:
         would_exceed_page = does_collection_contain_more_than_n_matching_documents(
-            collection=mdb[collection_name], filter_=filter_, n=limit
+            collection=mdb[collection_name], filter_=filter_, n=max_page_size
         )
         if not would_exceed_page:
             will_paginate = False
@@ -179,7 +179,7 @@ def list_resources(req: ListRequest, mdb: MongoDatabase, collection_name: str):
             mdb[collection_name].find(
                 filter=filter_,
                 projection=projection,
-                limit=limit,
+                limit=max_page_size,
                 sort=[(id_field, 1)],
                 allow_disk_use=True,
             )
--- a/nmdc_runtime/api/endpoints/util.py
+++ b/nmdc_runtime/api/endpoints/util.py
@@ -151,14 +151,23 @@ def list_resources(req: ListRequest, mdb: MongoDatabase, collection_name: str):
         else:
             filter_ = merge(filter_, {id_field: {"$gt": last_id}})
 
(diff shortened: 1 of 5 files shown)


eecavanna:
I just noticed Mongo's `collection.countDocuments` function accepts a `limit` option.

https://www.mongodb.com/docs/manual/reference/method/db.collection.countDocuments/#syntax

If pymongo's `collection.count_documents` function supports that, then we could use that instead and avoid having this helper function I am introducing in this PR (that could be addressed now or via a future, refactoring PR).

https://pymongo.readthedocs.io/en/stable/api/pymongo/collection.html#pymongo.collection.Collection.count_documents

nmdc_runtime/api/endpoints/util.py
```diff
@@ -114,13 +126,23 @@ def list_resources(req: ListRequest, mdb: MongoDatabase, collection_name: str):
...
-    # TODO: Since we don't need to know the exact count, but we do need to know whether
-    #       it would be greater than `limit`; consider using an aggregation pipeline to
-    #       count only up to "limit + 1" (see: https://stackoverflow.com/a/67503437).
+    # Note: We will paginate them unless either:
+    #       - the `max_page_size` is not a positive integer
+    #       - the number of documents matching the filter does not exceed `max_page_size`
     #
-    if (limit == 0) or (mdb[collection_name].count_documents(filter=filter_) <= limit):
+    will_paginate = True
+    if not isinstance(max_page_size, int):
```
dwinston, line 136: alternate:
```python
if not max_page_size:
    will_paginate = False
```

PR #1174, opened 2025-08-27
Enable referential integrity checking within `perform_mongo_updates` op

### LINKED ISSUE #969: Add referential integrity check to `apply_metadata_in` Dagster job
Add real-time referential integrity checking to the `apply_metadata_in` Dagster job.

## Relevant code

https://github.com/microbiomedata/nmdc-runtime/blob/d380b25e6d5a881332beb6dabb921128f6228610/nmdc_runtime/site/graphs.py#L129-L133

The `perform_mongo_updates` function uses `validate_json` under the hood:

https://github.com/microbiomedata/nmdc-runtime/blob/d380b25e6d5a881332beb6dabb921128f6228610/nmdc_runtime/site/ops.py#L537-L539

Nowadays, the `validate_json` function has a Boolean parameter that enables real-time referential integrity checking. See the function's docstring for details.

eecavanna:
On this branch, I added real-time referential checking to the `apply_metadata_in` Dagster graph.

### Details

Specifically, I:
- Added the `check_inter_document_references=True` kwarg to a call to `validate_json()` so that the `apply_metadata_in` Dagster graph performs referential integrity checking (in addition to the schema validation it already performed).
- Updated the existing `test_apply_metadata_in_functional_annotation_agg` test so that the document _referenced by_ the documents the test inserts via the Dagster op, exists. That way, the insertion does not fail due to the newly-introduced referential integrity check.
- Added a new test, named `test_apply_metadata_in_checks_referential_integrity`, which demonstrates that, when a document being inserted via the Dagster op includes a broken reference, the document does not get inserted.
- Added some `TODO`s about documenting undocumented functions.

The generically-named `perform_mongo_updates` function is only called by the `apply_metadata_in` graph.

### Related issue(s)

Fixes #969 

### Related subsystem(s)

- [ ] Minter
- [x] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by...

- Adding a test that demonstrates the new functionality
- Ensuring all tests pass locally and on GHA

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/tests/test_ops/test_ops.py
+++ b/tests/test_ops/test_ops.py
@@ -104,3 +104,45 @@ def test_apply_metadata_in_functional_annotation_agg(op_context):
     db.workflow_execution_set.delete_many({"id": workflow_execution_id})
     for doc_spec in docs["functional_annotation_agg"]:
         db.functional_annotation_agg.delete_many(doc_spec)
+
+
+def test_apply_metadata_in_checks_referential_integrity(op_context):
+    db = op_context.resources.mongo.db  # concise alias
+    
+    # Confirm the _referenced_ document does not exist (anywhere the schema says it can).
+    workflow_execution_id = "nmdc:wfmtan-00-000000.1"
+    assert db.workflow_execution_set.count_documents({"id": workflow_execution_id}) == 0
+
+    functional_annotation_agg_member = {
+        "was_generated_by": workflow_execution_id,
+        "gene_function_id": "KEGG.ORTHOLOGY:K00005",
+        "count": 10,
+        "type": "nmdc:FunctionalAnnotationAggMember",
+    }
+
+    docs = {"functional_annotation_agg": [functional_annotation_agg_member]}
+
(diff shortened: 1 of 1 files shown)


dwinston (approved):

eecavanna:
Thanks, @dwinston. Because it can update, I don't think the solution I've implemented in this PR will suffice. It covers insertions, but not updates.

For example: if the caller wants to update an existing document (matched by its `id`) so it has a different `type`, the solution I implemented here would _not_ check references that were already pointing to that document (from documents not involved in the operation) to confirm they are still schema-compliant references.

There is a function that can be used to check referential integrity in this case (it uses a MongoDB transaction to do the writes, check references, and then roll back the writes), but it it not what I did here and may complicate this function enough that we defer adding referential integrity checking to this function until after Q4 ends (which was always an option for the three referential integrity-related tickets on the squad board).

I'll think about this more next week, and we can discuss it at the next squad meeting.

hesspnnl (approved):

Issue #1196, opened 2025-09-06
Some Runtime tests are being skipped


eecavanna:
The Runtime's test suite include some (5) tests that are being skipped via `@pytest.mark.skip`. You can search the codebase for `@pytest.mark.skip(` (notice the trailing open parenthesis) in order to identify them.

They are:
- [x] A: `test_nmdc_jsonschema_using_new_id_scheme` in `tests/test_util.py`
- [x] B: `test_update_operation` in `tests/test_api/test_endpoints.py`
- [ ] C: `test_update_01` in `tests/test_api/test_metadata.py`
- [ ] D: `test_schema_conformance` in `tests/test_data/test_integrity.py`
- [ ] E: `test_ensure_jobs` in `tessts/test_graphs/test_ensure_jobs.py`

This ticket represents the task of un-skipping those tests and fixing the codebase so that those tests pass.

eecavanna:
I tried unskipping each of the five tests.

Two of the tests passed out-of-the-box (although one of them left a Mongo connection open, so I added a `close()` statement to it).

One of the tests failed because an ID pattern in the schema didn't have the prefix the test was checking for. I updated the test to check for that additional prefix and the test passed. I want the author of the test to review that with respect to what they designed the test to do.

Two of the still tests fail when un-skipped. For those, I updated their `reason` strings (the reason they are being skipped) with the error message from pytest.

eecavanna:
Skipping tests is mentioned in https://github.com/microbiomedata/nmdc-runtime/issues/467.

PR #1197, opened 2025-09-06
Un-skip tests that were being skipped

eecavanna:
On this branch, I tried unskipping each of the five tests that were being unconditionally skipped.

They are:
- [x] A: `test_nmdc_jsonschema_using_new_id_scheme` in `tests/test_util.py`
- [x] B: `test_update_operation` in `tests/test_api/test_endpoints.py`
- [ ] C: `test_update_01` in `tests/test_api/test_metadata.py`
- [x] D: `test_schema_conformance` in `tests/test_data/test_integrity.py`
- [ ] E: `test_ensure_jobs` in `tessts/test_graphs/test_ensure_jobs.py`

Two of the tests passed out-of-the-box (although one of them left a Mongo connection open, so I added a close() statement to it).

One of the tests failed because an ID pattern in the schema didn't have the prefix the test was checking for. I updated the test to check for that additional prefix and the test passed. I want the author of the test to review that with respect to what they designed the test to do.

Two of the still tests fail when un-skipped. For those, I updated their reason strings (the reason they are being skipped) with the error message from pytest.

Will fix: https://github.com/microbiomedata/nmdc-runtime/issues/1196

### DIFF
--- a/tests/test_api/test_endpoints.py
+++ b/tests/test_api/test_endpoints.py
@@ -121,7 +121,6 @@ def ensure_test_resources(mdb):
     }
 
 
-@pytest.mark.skip(reason="Skipping because test causes suite to hang")
 def test_update_operation():
     mdb = get_mongo(run_config_frozen__normal_env).db
     rs = ensure_test_resources(mdb)
--- a/tests/test_api/test_metadata.py
+++ b/tests/test_api/test_metadata.py
@@ -115,7 +115,9 @@ def test_changesheet_update_slot_with_range_uriorcurie():
         mdb.study_set.delete_one({"id": "nmdc:" + local_id})
 
 
-@pytest.mark.skip(reason="no /site-packages/nmdc_schema/external_identifiers.yaml ?")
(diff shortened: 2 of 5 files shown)


eecavanna:
Hi @dwinston, I want to discuss these tests that were being skipped, during one of our upcoming pairing sessions. I took a pass at un-skipping them tonight, and this is as far as I got.

dwinston:
> Quick way to run one of these tests:
> 
> ```shell
> # On host:
> make up-test
> docker compose -f docker-compose.test.yml exec test bash
> 
> # In container (replace path and function name as needed) (re-run as needed):
> uv run --active pytest tests/test_graphs/test_ensure_jobs.py -k test_ensure_jobs
> ```
> 

Thanks. I have been in the habit of doing e.g.
```shell
make down-test # reset database
make up-test
make run-test ARGS="-k 'test_ensure_jobs'"
```
, which I suppose adds some overhead.

dwinston (dismissed):

PR #1284, opened 2025-10-16
Pin Swagger UI CSS and JS versions to `5.29.4`

### LINKED ISSUE #1283: Pin Swagger UI version
...to avoid situations like this: https://github.com/microbiomedata/nmdc-runtime/issues/1273

eecavanna:
On this branch, I pinned the version of Swagger UI that we are using, to the latest available version.

### Details

By default, FastAPI pins it to version 5, but allows for the minor and patch versions to change over time.

Since we recently introduced JavaScript code that accesses/manipulates some "internal" HTML elements of Swagger UI, what the Swagger UI developers consider to be a "breaking change" is different (i.e. more narrow) than what we consider to be a "breaking change." By pinning the version, that difference does not come into play.

### Related issue(s)

Fixes #1283 

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [x] Other

Swagger UI only.

### Testing

I tested these changes by visiting Swagger UI locally, and visiting both of those `.js` and `.css` URLs directly. When visiting Swagger UI locally, I disabled caching via DevTools, reloaded the page a bunch of times, and confirmed the URLs from which the web browser was fetching the JS and CSS files were what I expected.

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/nmdc_runtime/api/main.py
+++ b/nmdc_runtime/api/main.py
@@ -361,7 +361,14 @@ def custom_swagger_ui_html(
             "onComplete": f"""<unquote-safe>() => {{ {onComplete}; dispatchEvent(new Event('nmdcInit')); }}</unquote-safe>""",
         }
     )
+    # Pin the Swagger UI version to avoid breaking changes (to our own JavaScript code that depends on Swagger UI internals).
+    # Note: version `5.29.4` was released on October 10, 2025.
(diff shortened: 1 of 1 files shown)


shreddd:
Do we also need to add this to a JavaScript / npm based manifest somwhere to allow dependabot to test for updates?

PR #1318, opened 2025-11-06
Fix bugs where response validation would fail when projection was used

### LINKED ISSUE #1317: projection not working on the list jobs endpoint
**Describe the bug**
Projection is not working on the list jobs endpoint. This is not working on both dev and prod

**To Reproduce**
curl -X 'GET' \
  'https://api-dev.microbiomedata.org/jobs?filter=%7B%22config.was_informed_by%22%3A%22nmdc%3Adgns-12-w4reqz25%22%7D&max_page_size=400&projection=workflow.id' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJjbGllbnQ6c3lzMHdtNjYiLCJleHAiOjE3NjI0NTc0MzZ9.xFa9E7Hy9mafFP_aJ95O0A8DvzDn1S3wzIga70FMbD0'

returns a response code of 500 with the swagger UI

Error: response status is 500

**Expected behavior**
I expect projects to work as they do on other endpoints. For example a filter plus projection on the nmdcschema list from collection endpoint returns a 200.
curl -X 'GET' \
  'https://api-dev.microbiomedata.org/nmdcschema/data_object_set?filter=%7B%22id%22%3A%22nmdc%3Adobj-11-7fg14479%22%7D&max_page_size=20&projection=url' \
  -H 'accept: application/json'

**Acceptance Criteria**
A 200 response code when adding a projection on the jobs endpoint

cc @eecavanna @shreddd 

I do not consider this issue a blocker for the release on Monday.

hesspnnl:
On this branch, I investigated the bug in the get /jobs endpoint when trying to input a projection. 

### Details

I fixed the response object and am now stripping the _id mongo given field. 
### Related issue(s)

fixes #1317 
### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by run the dev stack locally and calling the endpoint with projections

### Documentation

- [ ] Other (explain below)

### Maintainability

- [ ] The PR title is in the imperative mood (e.g. "Do X") and not the declarative mood (e.g. "Does X" or "Did X")

### DIFF
--- a/nmdc_runtime/api/endpoints/jobs.py
+++ b/nmdc_runtime/api/endpoints/jobs.py
@@ -7,7 +7,6 @@ from pymongo.database import Database
 from fastapi import APIRouter, Depends, Query, HTTPException, Path
 from pymongo.errors import ConnectionFailure, OperationFailure
 from starlette import status
-from nmdc_runtime.api.models.metadata import Doc
 from nmdc_runtime.api.core.util import (
     raise404_if_none,
 )
@@ -15,6 +14,7 @@ from nmdc_runtime.api.db.mongo import get_mongo_db
 from nmdc_runtime.api.core.idgen import generate_one_id
 from nmdc_runtime.api.endpoints.util import list_resources, _claim_job, strip_oid
 from nmdc_runtime.api.models.job import Job, JobClaim, JobIn
+from nmdc_runtime.api.models.metadata import Doc
 from nmdc_runtime.api.models.operation import Operation, MetadataT
 from nmdc_runtime.api.models.site import (
     Site,
@@ -26,6 +26,9 @@ from nmdc_runtime.api.models.util import ListRequest, ListResponse, ResultT
 router = APIRouter()
 
 
+# Note: We use the generic `Doc` class—instead of the `Job` class—to describe the response
+#       because this endpoint (via `ListRequest`) supports projection, which can be used to omit
+#       fields from the response, even fields the `Job` class says are required.
 @router.get(
     "/jobs", response_model=ListResponse[Doc], response_model_exclude_unset=True
 )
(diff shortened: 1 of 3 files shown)


hesspnnl:
@eecavanna This same bug will also impact any other endpoint that follows this pattern:
`response_model=ListResponse[object_model]`
If the projection does not include all "required" fields on the object_model, a pydantic error occurs because the model is technically invalid. 
Following this generic pattern from the nmdcschema.py router allows the responses to not include all required fields while also following the ListResponse pattern:
`response_model=ListResponse[Doc]`

eecavanna:
Thanks for implementing this! I'll review it now.

eecavanna:
Here's a copy/paste from my CLI, showing the "**before**" behavior (i.e. the bug). 

#### Lacking a projection: PASS

```console
$ curl -X GET 'https://api.microbiomedata.org/objects?max_page_size=1'
{"resources":[...],"next_page_token":"..."}
```

#### Having a projection: FAIL

```console
$ curl -X GET 'https://api.microbiomedata.org/objects?max_page_size=1&projection=name'
Internal Server Error%
```

eecavanna:
The newly-referenced Pydantic class is `Doc`, which is defined like this (in `nmdc_runtime/api/models/metadata.py`). It's a model that can have _any_ fields.

```py
class Doc(BaseModel):
    model_config = ConfigDict(extra="allow")
```

This makes validation pass and fixes the bug.

The tradeoff is that the response documentation becomes generic (it looks like this).

image

That is in contrast to the "before" response documentation, which was more specific, but could be inaccurate when using projection (this was the bug).

image

I will add some `TODO` comments to the code, to remind developers that there is an opportunity to reintroduce more specific-ness (if we can find a way to do so in a maintainable way).

PR #1343, opened 2025-12-04
Document site client credential customization procedures for local development

### LINKED ISSUE #1329: Update development documentation to explicitly explain the (local) site client creation process
This ticket is a result of the documented issues (#1319 and PR #1320) a developer had when trying to test local site client access to the `/queries:run` endpoint. After implementing a quick fix for the developer, Eric and I chatted about what is the best way to not impact the production environment, but still allow developers to customize their local dev environment. We decided on updating the documentation to be more explicit on instructions for this process. Here are some notes to go off of:

- There is currently a step that says "Ensure you have a .env file..." I can see this new section fitting well immediately below that; for example: Note: Some environment variables have a special purpose during the first boot of the Runtime; specifically, API_SITE_CLIENT_ID and API_SITE_CLIENT_SECRET, which dictate the credentials for the base site's site client. If you want to change those values after the first boot, you can do so by: (1) turning off the Runtime, (2) deleting the base site from the Mongo database, (3) changing those values in the .env file, and (4) turning on the Runtime. This can come in handy when you want to use specific site client credentials.
- [OH] : I think from the documentation it reads (at least it did to me) that we could run openssl rand -hex 32 to generate random strings that would be read from those env variables and created in my local DB for testing.[EC] : That is the case, but the values have to be in place before the first boot (which is not stated in the docs or .env.example file
- add the guidance as a comment to the .env.example file (like lines [47-49](https://github.com/microbiomedata/nmdc-runtime/blob/a2fbd06233896b03eecd0d224fb16c21c377a3a5/.env.example#L47-L49) and [58-60](https://github.com/microbiomedata/nmdc-runtime/blob/a2fbd06233896b03eecd0d224fb16c21c377a3a5/.env.example#L58-L60), for example) and to the local development section of the README. 
- [EC] Here is a workaround that people can employ when they are in that situation; i.e. they have a specific client_id and client_secret that they want to use, but they did not have those in place in the .env file before first booting the Runtime. The workaround is:
```
Stop the Runtime
Using a Mongo client, go to the sites collection and delete the document whose id matches the value of the API_SITE_ID environment variable
Customize the API_SITE_CLIENT_ID and API_SITE_CLIENT_SECRET environment variables
Start the Runtime
The Runtime will then create the site, and it will have a site client having the specified client_id and client_secret.
```

hesspnnl:
On this branch, I added documentation on creating a custom local site client after the first boot of the dev runtime stack.

### Related issue(s)

Fixes #1329 

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by...

### Documentation

- [x] I have **updated** all relevant documentation so it will remain accurate
- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ http://127.0.0.1:8000/redoc/.
 
 Some environment variables in the `.env` file are only processed during the first boot of the FastAPI container; e.g., the `API_SITE_CLIENT_ID` and `API_SITE_CLIENT_SECRET` environment variables, which dictate the credentials of the base site's site client.
 
-To people that want the base site's site client credentials to have specific values (e.g., to match something in a dependent application), we recommend customizing those environment variables **before** starting up the FastAPI container.
+To people who want the base site's site client credentials to have specific values (e.g., to match something in a dependent application), we recommend customizing those environment variables **before** starting up the FastAPI container.
 
 In case you have already started up the FastAPI container (this is common), all is not lost! You can still customize the credentials—here's how:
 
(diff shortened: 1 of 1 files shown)


eecavanna:
Thanks, @hesspnnl! I'll review this before Monday.

PR #1389, opened 2026-02-12
Make `GET /data_objects/study/{study_id}` faster and allow `DataObject` repetition

### LINKED ISSUE #1373: API endpoint `/data_objects/study/{study_id}` returns HTTP 502 Bad Gateway (prod)
I'm having trouble with this request timing out via postman 2

https://api.microbiomedata.org/data_objects/study/nmdc%3Asty-11-34xj1150

I'm getting at  502: Bad gateway

eecavanna:
On this branch, I resolved some performance issues and a logical issue affecting the `GET /data_objects/study/{study_id}` API endpoint.

When running locally, the endpoint now takes only 30 seconds to respond to the following request instead of 220 seconds (about 7x faster).

```shell
curl 'https://api.microbiomedata.org/data_objects/study/nmdc%3Asty-11-34xj1150'
```

#### A. Performance issues

As described in #1373, a specific request was timing out in production. Locally, the same request was taking 3.5 minutes to get a response. On this branch, I fixed this by doing two things. First, I switched the "find all biosamples associated with this study" step from using the "linked instances" plumbing to using a plain Mongo query involving the `Biosample.associated_studies` field. Then, I changed the way the endpoint uses the "linked instances" plumbing—I changed it from requesting downstream `DataObjects` for all 5000+ `Biosample`s at once, to requesting them for batches of 1000 `Biosample`s each and then merging the results later. As a result, the same request is taking 25-35 seconds to get a response (locally). I expect it to be faster in dev/prod.

I also introduced a `duration_logger` context manager, which developers can use with `with ...` in order to log the durations of tasks. We recently introduced this into `nmdc-server` also ([here](https://github.com/microbiomedata/nmdc-server/blob/174d9457355d12ba6489e195bcd6a4e51fca2eb9/nmdc_server/ingest/common.py#L103)), and I have found it helpful when trying to identify performance issues in both applications.

#### B. Logical issue

In dev/prod, the endpoint's response currently includes all the requested `DataObject`s, but it only includes each one **at most once**. If a given `DataObject` is downstream from more than one `Biosample`, it will only be present in _one_ of those `Biosample`s' lists in the response. While this avoids having multiple copies of the entire `DataObject` in the response (which can save bandwidth and reduce response time), it contradicts the endpoint's user-facing description. On this branch, I made it so each `DataObject` appears in the list of **each** `Biosample` that is upstream from it.

### Related issue(s)

Fixes #1373 

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by (a) ensuring all GHA workflows pass, and (b) manually issuing the original request shown in the linked issue many times locally. I also did performance testing, which you can learn more about in the linked issue (specifically, [this comment](https://github.com/microbiomedata/nmdc-runtime/issues/1373#issuecomment-3888345646)).

Note: In my opinion, the test suite targeting this endpoint is due for a refactor.

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/nmdc_runtime/api/endpoints/find.py
+++ b/nmdc_runtime/api/endpoints/find.py
@@ -287,7 +287,7 @@ def find_data_objects_for_study(
     num_biosample_ids = len(biosample_ids)
     logging.debug(f"Found {num_biosample_ids} Biosamples.")
 
-    # Divide the `Biosample` IDs into batches.
+    # Divide the `Biosample` IDs into batches (returning early if there are no such IDs).
     #
     # Note: This is a performance optimization of the `get_linked_instances` function usage below.
     #       In our (local) testing with the 5260 `Biosample`s associated with the `Study` whose ID
@@ -298,6 +298,8 @@ def find_data_objects_for_study(
     #
     num_ids_per_batch = 1000  # this can be "tuned"
     biosample_id_batches = []
+    if num_biosample_ids == 0:
+        return biosample_data_objects  # no need to proceed further if there are no biosample IDs
     if num_biosample_ids <= num_ids_per_batch:
         biosample_id_batches = [biosample_ids]  # a single batch
     else:


eecavanna:
Yes, I'll test it now and report back. Thanks for providing that snippet. (I will be away from my desk from about 1:10-1:20pm, and I don't think I'll have a result before then.)

PR #1430, opened 2026-04-11
Create unique compound index on `(name, associated_studies)` in `biosample_set` collection

### LINKED ISSUE #1417: Make a compound unique key constraint in mongo for biosamples
**Is your feature request related to a problem? Please describe.**
on biosample set a compound key of name + associated-studies should be unique

**Screenshots**

**Describe the solution you'd like**
This is to prevent duplicate sample names with a given study, see https://github.com/microbiomedata/nmdc-schema/issues/2878

Example user story template:
AS A {user|persona|system},
[INSTEAD OF {current condition}]
I WANT TO {action} [IN {mode} TIME | IN {differentiating performance units} TO {utility performance units}]
[SO THAT {value or justification}]
[NO LATER THAN {best by date}]

**Acceptance Criteria**
- Submissions to nmdc-runtime post endpoints that do not have a unique value for name + associated studies for biosample_set records return errors.
- this should be true for new submissions (via json:submit) and through endpoints which modify records like queries:run and changesheets.

Example scenario-based template:
Given (some given context or precondition), when (I take this action), then (this will be the specific outcome).

**Describe alternatives you've considered**
Not enforcing this may lead to duplicate sample names in mongo which then requires subsequent cleanup.

**Additional context**
- dependent on 
   - [x] https://github.com/microbiomedata/nmdc-schema/pull/2901

eecavanna:
On this branch, I updated the Runtime's bootup routine to ensure the Mongo database has a unique compound index on `(name, associated_studies)`, so that no two biosamples associated with the same study can have the same name.

### Details

I also implemented a generic test that demonstrates that such a unique compound index works the way I expect.

### Related issue(s)

Fixes #1417

### Related subsystem(s)

- [ ] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I tested these changes by confirming all tests pass, including the newly-introduced one.

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/nmdc_runtime/api/core/metadata.py
+++ b/nmdc_runtime/api/core/metadata.py
@@ -759,7 +759,7 @@ def _validate_changesheet(df_change: pd.DataFrame, mdb: MongoDatabase):
             # TODO: Apply the update commands to the "nmdc" MongoDB database within a transaction
             #       and return an error to the user if there are any errors. At this point, if we
             #       are going to perform the commands on the real database within a transaction for
-            #       validation purposes, we _may as well_ perform it for reals, without Dagster! 
+            #       validation purposes, we _may as well_ perform it for reals, without Dagster!
             pass
 
     validation_errors = []
--- a/nmdc_runtime/api/core/metadata.py
+++ b/nmdc_runtime/api/core/metadata.py
@@ -599,6 +599,11 @@ def mongo_update_command_for(df_change: pds.DataFrame) -> Dict[str, list]:
     """
     Creates a dictionary of update commands to be executed against the Mongo database.
 
+    FIXME: The "Returns" section of this docstring is inaccurate. The return value's type hint
+           is also inaccurate. See how the return value is built within the function body to
+           get idea of the real data type and conceptual content of the value. Either fix this
+           or eliminate changesheet support altogether.
+
     Parameters
     ----------
     df_change : pds.DataFrame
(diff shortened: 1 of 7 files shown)


aclum:
I see this covers json:submit and you can't change an ID with changesheets. What about queries:run updates?

eecavanna:
Maybe needless to say: Once the index exists in Mongo, Mongo, itself, will prevent operations that violate the index.

The things I'm adding to this branch are about UX (user experience). Since some writes happen via Dagster, the endpoint (and user) doesn't have a way to know whether the write succeeds or fails (unless they follow up with Dagster). Currently, the endpoint will say "yes, I've created the job to perform the write," whether that write will eventually succeed or fail. So, I'm updating the endpoints to check whether the write _would_ (if performed _right now_) fail, so the endpoint can return an actionable response to the user.

aclum (dismissed):

eecavanna:
For now, only `/metadata/json:submit` will return an error response when the requested change would violate the uniqueness index. I added `TODO` comments about updating `/metadata/changesheets:validate` and `/metadata/changesheets:submit` (people can introduce duplicate `name` values through that endpoint) and `/queries:run` to do the same. I'll file follow-on tickets about those.

In short, once I merge this PR into `main`:

1. The MongoDB index will prohibit multiple biosamples associated with the same study from having the same name as one another.
2. When someone submits something via `/metadata/json:submit` that would violate that, the API will return an error response. This PR contains a test demonstrating that.
3. When someone submits something via either `/metadata/changesheets:validate`, `/metadata/changesheets:submit`, or `/queries:run` that would violate that, the API will not return an error response, but the downstream write performed by Dagster will fail (due to item 1 above).
4. I'll create two follow-on tickets: one about `/metadata/changesheets:validate/` and `/metadata/changesheets:submit` and one about `/queries:run`
   - One ticket is: https://github.com/microbiomedata/nmdc-runtime/issues/1442
   - The other ticket is: https://github.com/microbiomedata/nmdc-runtime/issues/1443

PR #1526, opened 2026-06-09
Update minter to mint dot-integer-suffixed IDs for workflow executions

### LINKED ISSUE #529: Mint & Validate versioned `WorkflowExecutionActivity` `nmdc:` IDs for workflows of same type and `OmicsProcessing`
Endpoints that accept post (`/workflows_activities`, `json:submit`) for `WorkflowExecutionActivity` subclasses should make sure versioning rules are being followed. The schema is validating the syntax but not that the incrementation is correct. 

Expected behavior: The first time a workflow runs (unique value for `was_informed_by`)a workflow example identifier would be nmdc:wfrqc-11-abc1d.1, the second time a workflow is run for the same value of `was_informed_by` **the workflow should keep the identifier through the blade and increment the ID version** (ie nmdc:wfrqc-11-abc1d.2)

There is currently no validation on this and we have instances where a second run of a workflow for a value of `was_informed_by` mints a new ID instead of incrementing the version id.

cc @shreddd to identify someone to work on this.

eecavanna:
On this branch, I implemented a new minter endpoint (at `POST /pids/mint/workflow_execution_id`), which people can use to mint "dot integer"-suffixed IDs for concrete subclasses of `WorkflowExecution`.

image

image

### Details

Previously, the minter would not mint ID values ending with "dot integer" suffixes (e.g. `nmdc:wfnom-11-123123.1`).

However, workflow automation developers wanted the `id` values of workflow executions to be "versioned" (i.e. to share a common base, but to end with a "dot integer" suffix; e.g. `.1`, `.2`, etc.), which the schema _does_ allow. So, workflow automation developers would append those suffixes locally and then submit them to the Runtime, resulting in the use of non-minted IDs in production Mongo data.

This new endpoint can be used to get _minted_, "versioned" IDs. The way it works is:
- The client specifies the schema class (using its `class_uri`, as is the case with the original `/pids/mint` endpoint) and, optionally, an existing WFE ID
- The server will mint an ID via one of these two "code paths":
  - A) If the client specified an existing WFE ID, the minted ID will have the same base portion, and the "dot integer" portion will be 1 greater than the highest one currently in use (e.g. if `.1` and `.3` are in use, the minted ID will end in `.4`, not `.2`)
  - B) If the client didn't specify an existing WFE ID, the minted ID will have some unique base portion, and the "dot integer" portion will be `.1`. In this case, _two_ IDs will actually be minted, although only the "dot integer"-suffixed one will be returned to the client (the base one will not)
- Determining whether an ID is "in use" involves checking _both_ the minter's own history  (since we would be minting it) _and_ the `workflow_execution_set` collection (since people can use IDs minted by different minter instances, and even contrived IDs as long as they are schema compliant)

I also modularized some existing minter code, and made it so the minter ensures the `minter.id_records` collection has a unique index on its `id` field (which it does in production, but I assume that was created manually since I don't see any code on `nmdc-runtime/main` that does it).

### Related issue(s)

Fixes #529

### Related subsystem(s)

- [x] Minter
- [ ] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I added tests that exercise the new functionality and confirmed they pass. I also confirmed all pre-existing tests still pass. This is the case both locally and on GHA.

### Documentation

- [ ] Other (explain below)

### Maintainability


eecavanna:
The test suite is failing on GHA because the test stack isn't starting up.

```
wait-for-it.sh: waiting 300 seconds for fastapi:8000
wait-for-it.sh: timeout occurred after waiting 300 seconds for fastapi:8000
wait-for-it.sh: strict mode, refusing to execute subprocess
make: *** [Makefile:53: run-test] Error 124
```

It starts up locally and, indeed, all tests pass locally.

---

Turns out Git was configured to ignore the files in `lib/`, so there were a few files that only existed in my local environment and not on this branch (oops). Fixed now.

eecavanna:
@codex darling sweet honey babe, will you review this pull request?

eecavanna:
Note to reviewers:
- I don't expect you to scrutinize the diff of this PR
- I do want you to read the endpoint behavior in the PR description

As an alternative (or supplement) to a PR review, we could merge it into `main` and kick the tires on it in dev.

I have had Codex "review" it (locally) and it came back with no "concerns".

kheal:
@eecavanna - Is the expectation that all WFE ids be minted with this endpoint rather than the /mint endpoint (which currently can mint the _base_ of a WFE record)?  When we (@bmeluch and myself) run these workflows, we often want to mint hundreds of IDs.  We've always used the /mint endpoint to grab a batch of ids to make it more efficient (and then appended the .1 manually for new workflows).

For 'reruns' the functionality in this PR is good to have, but it's limiting for new workflows where we might want to mint many .1 appended ids at once.

eecavanna:
Thanks, @kheal!

> Is the expectation that all WFE ids be minted with this endpoint rather than the /mint endpoint (which currently can mint the base of a WFE record)?

I **was** hoping that, once this endpoint was introduced, all future WFE IDs (for as long as people require them to end with `.INTEGER` suffixes) would, indeed, be minted with this endpoint.

> we often want to mint hundreds of IDs. We've always used the /mint endpoint to grab a batch of ids to make it more efficient (and then appended the .1 manually for new workflows).
>
> [...]
>
> for new workflows where we might want to mint many .1 appended ids at once.

Thanks for pointing this out. With this in mind, I plan to add that ability to either this new endpoint or a second new endpoint (with this one _maybe_ being renamed). I'll do it in this PR. I'll tag your GitHub username in a follow-up comment here once it's been implemented.

kheal:
@eecavanna is it possible to add the .1 integer for WFEs to the exisiting mint endpoint? It makes it much easier for us (again, @bmeluch  and myself) to use the same endpoint when minting ids (again, for _new_ WFEs only) rather than a separate one (see here for code we'd need to reconfigure if we use more than one endpoint for this: https://github.com/microbiomedata/nmdc_mass_spectrometry_metadata_generation/blob/main/nmdc_ms_metadata_gen/id_pool.py .  

The WFE ids currently minted by the minter _do not adhere to the schema_ so I think simply adding a .1 on runtime to those would be a more straightforward approach. Otherwise the minter should not mint WFEs at all if the expectation is to use these minted ids.

eecavanna:
> The WFE ids currently minted by the minter do not adhere to the schema

Yikes! I didn't realize that. I confirmed it just now:

image

> References: [nmdc_materialized_patterns.schema.json](https://microbiomedata.github.io/nmdc-schema/downloads/nmdc_materialized_patterns.schema.json) (I then unescaped `\\.` to `\.`); and https://regexr.com/

With this in mind, I actually do want to make changes to the `/mint` endpoint. I had assumed all the `id`s it was minting were schema-compliant. As you pointed out, they aren't. I had been trying to avoid touching that endpoint (to reduce risk), but I do think it's necessary now. So, I'm going to overhaul this PR.

Thanks for pointing this out!

eecavanna:
I created ticket https://github.com/microbiomedata/nmdc-runtime/issues/1528 just now, about the fact that the `id` values returned by the `/pids/mint` endpoint are **not schema compliant**.

**Next steps:** Off the top of my head, I think I will (a) update the `/pids/mint` endpoint to include `.1` at the end of the `id` values it generates for concrete subclasses of `WorkflowExecution`, so that those `id`s are **schema compliant**. People would still be able to use that to generate "hundreds of IDs" per request. In addition, I think I will (b) introduce a new endpoint (likely repurposing this `/pids/mint/workflow_execution_id` one, maybe renaming it slightly) that handles the incrementing of the "dot integer" suffix, so that _that_ `id` (ending in `.2`, `.3`, etc.) is a **minted** `id`. That endpoint will only generate a single `id` per request.

PR #1559, opened 2026-07-17
Configure Dagster to limit `ensure_alldocs` to 45 mins and to skip launching new runs when already running

### LINKED ISSUE #1356: Limit concurrency of `materialize_alldocs` Dagster op
While working on ticket https://github.com/microbiomedata/nmdc-runtime/issues/1165, it seemed to me that having multiple instances of the `materialize_alldocs` op running simultaneously would—at least sometimes–lead to at least some of them not finishing.

Indeed, after having subsequently removed that op from a few upstream graphs—even though we simultaneously increased the frequency at which the containing `ensure_alldocs` graph runs (we changed it from daily to hourly)—I have not seen that symptom arise.

It is still possible that, in the future, a developer could reintroduce the `materialize_alldocs` op into some other graph, leading to multiple instances of the op running simultaneously again. We may be able to mitigate that by limiting the concurrency of the op.

This ticket represents the following tasks:
- [ ] Determine whether Dagster allows us to limit the concurrency of an op
- [ ] If we can, limit the concurrency of the `materialize_alldocs` op to 1

eecavanna:
On this branch, I made changes to the way Dagster manages the `ensure_alldocs` job.

#### Limiting run duration

Previously, the job could run indefinitely (that is true about all Dagster jobs by default). Normally, that wasn't a problem, since this job normally takes about 20 minutes to run. During our monthly releases (and at other times where we shut down Dagster), this was leading to multiple runs of the job "clogging" the queue, preventing unrelated jobs from being run (such as those that apply changesheets). That phenomenon is documented in the [Dagster docs](https://docs.dagster.io/deployment/execution/run-monitoring#detecting-run-worker-crashes):

image

On this branch, I enabled Dagster's "[run monitoring](https://docs.dagster.io/deployment/execution/run-monitoring)" feature and configured it to **"fail" any `ensure_alldocs` runs that don't finish within 45 minutes**. For reference, the runs of that job normally finish in 20 minutes.

#### Skipping if already running/queued

Also, previously, Dagster was configured to schedule a run of the `ensure_alldocs` jobs every hour, _even if_ there were other runs of the job still running or queued.

On this branch, I configured Dagster to only schedule a run of the hourly `ensure_alldocs` job **if there are no running or queued runs of it.** If there are already running or queued runs of it at the time, Dagster will skip the new run ~(and set its status to `SKIPPED`)~. Note that this does not apply to launching the `ensure_alldocs` job manually via the Dagster UI—it only applied to the hourly scheduled runs (although the "already running" instance can have been launched by either means).

### Related issue(s)

Fixes #1356

### Related subsystem(s)

- [ ] Minter
- [x] Dagster
- [ ] MongoDB migrations
- [ ] Other

### Testing

I will test these changes in dev.

### Documentation

- [ ] Other (explain below)

### Maintainability

### DIFF
--- a/nmdc_runtime/site/dagster.yaml
+++ b/nmdc_runtime/site/dagster.yaml
@@ -10,6 +10,12 @@ run_launcher:
   module: dagster.core.launcher
   class: DefaultRunLauncher
 
+# Enable Dagster's "run monitoring" feature, so Dagster will enforce the `MAX_RUNTIME_SECONDS_TAG`
+# duration limits that we define at the job level (e.g. we define one for the `ensure_alldocs` job).
+# Docs: https://docs.dagster.io/deployment/execution/run-monitoring#general-run-timeouts
+run_monitoring:
+  enabled: true
+
 run_storage:
   module: dagster_postgres.run_storage
   class: PostgresRunStorage
--- a/nmdc_runtime/site/repository.py
+++ b/nmdc_runtime/site/repository.py
@@ -159,7 +159,7 @@ ensure_alldocs_hourly = ScheduleDefinition(
             # durations exceed this limit. We enable Dagster's "run monitoring" feature
             # via `run_monitoring.enabled: true` in `nmdc_runtime/site/dagster.yaml`.
             MAX_RUNTIME_SECONDS_TAG: timedelta(minutes=45).total_seconds(),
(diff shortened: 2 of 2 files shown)


eecavanna:
@codex review

